PR #95, opened 2014-11-27
Using graceful-ncp instead of ncp

adam-lynch:
This helps with EMFILE errors. See [adam-lynch/graceful-ncp](https://github.com/adam-lynch/graceful-ncp).

This is related to https://github.com/mllrsohn/node-webkit-builder/pull/123

### DIFF
--- a/lib/copy.js
+++ b/lib/copy.js
@@ -1,6 +1,6 @@
 var fs = require('fs')
 var path = require('path')
-var ncp = require('ncp').ncp
+var ncp = require('graceful-ncp').ncp
 var mkdir = require('./mkdir')
 var create = require('./create')
 
--- a/lib/move.js
+++ b/lib/move.js
@@ -2,7 +2,7 @@
 //licensed under the BSD license: see https://github.com/andrewrk/node-mv/blob/master/package.json
 
 var fs = require('fs')
-var ncp = require('ncp').ncp
+var ncp = require('graceful-ncp').ncp
 var path = require('path')
 var rimraf = require('rimraf')
 var mkdirp = require('mkdirp')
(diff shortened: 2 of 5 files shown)


traviswimer:
Another possibility that might work better if the dependency is unacceptable is to just allow `fs` to be passed as a parameter. In other words something like this maybe:

```javascript
var gfs = require("graceful-fs");
var fs = require("fs-extra")( gfs );
```

I didn't actually look at the code to know if this specifically is reasonable, but it would be nice to have the option of the library of you choice through some process.

jprichardson:
@traviswimer `graceful-fs` is already a part of `fs-extra` and has been for awhile, see: https://github.com/jprichardson/node-fs-extra/blob/master/lib/index.js#L8 The concern is that `ncp`, the copy module that `fs-extra` uses does not use `graceful-fs` and won't accept any dependency modules. 

Either way, I think I'm going to fork `ncp` into `fs-extra` as the `ncp` guys have been pretty silent lately.

jprichardson:
Done. https://github.com/jprichardson/node-fs-extra/releases/tag/0.15.0

jprichardson:
Includes `graceful-fs` now as well.

adam-lynch:
Could you clarify what impact this has? What do you mean by "io.js only"? Is this a backwards-incompatibile change? What I mean is will I have to switch to io.js everywhere my code is ran for this to work? I haven't read into io.js enough yet probably.

jprichardson:
@adam-lynch `0.15.0` includes `ncp`, uses `graceful-fs`, patches `ncp` for support for `io.js`. You won't have to modify any of your code as it works on both node.js and io.js.

adam-lynch:
Thanks

PR #107, opened 2015-01-27
Check if dest path is an existing dir and copy or move source in it

kosmasgiannis:
Hi there, 

I think that it would be better when the destination path is an existing directory to copy or move the source file or directory in it instead of failing silently. 

If you agree with the idea please consider this pull request.

Giannis Kosmas


jprichardson:
Yes, I like this. Would you be willing to add a test case? Thank you!

kosmasgiannis:
Hi,

I will try to write some test cases and append them to this pull request.

Thanks

kosmasgiannis:
I have added some tests for copy copySync and move. I hope they are ok.

kosmasgiannis:
Hi there! 

Did you have any chance to review this PR? 

I 'd like to use this functionality in my project and I 'd like to know if I should include my fork or not. 

Thanks and sorry for the pressure. 

Cheers,
Giannis

jprichardson:
Unfortunately you're going to have to maintain your own fork for the time being. Unless you want to close this PR and resubmit smaller ones that will be easier to evaluate. By looking over https://github.com/kosmasgiannis/node-fs-extra/commit/8d601e71fc09a4ae8f2b3b9f94f9e9909beb3593 I'm not sure if this will affect this https://github.com/jprichardson/node-fs-extra/commit/1ee77c8a805eba5b99382a2591ff99667847c9c9. 

Overall, there's a lot going in here and it's tougher to review. It'd be super nice to have just one PR per change. The PR should include both the change/patch and a test for the change patch. That's it. Doing this will be the fastest way to getting your PRs accepted. As it stands now, I'm going to have to spend some significant time reviewing.

kosmasgiannis:
Hi again,
thanks for the prompt reply, I will close this one and provide a new all-in-one asap.

jprichardson:
Just to be clear... an "all-in-one" isn't what I'm looking for. I'm looking for multiple PRs, each PR that contains at minimum, two commits. One commit with the test. The following commit with the patch(s). Does that make sense? i.e. the first commit, the test that you added should fail, the second commit should make the test pass. This should be in one PR. Since you wanted to make the change for more than one method, it'd probably be appropriate to have more than one PR.

kosmasgiannis:
OK, got it...

jprichardson:
(Edited last response, I know that GH doesn't send out emails for edits... unless that's changed, but wanted you to see the edit)

jprichardson:
=)

PR #111, opened 2015-02-03
Moving a file into an existing directory 

kosmasgiannis:
I believe this time it is OK.

### DIFF
--- a/lib/move.js
+++ b/lib/move.js
@@ -18,11 +18,19 @@ function mv(source, dest, options, callback){
   var clobber = options.clobber !== false
   var limit = options.limit || 16
 
-  if (shouldMkdirp) {
-    mkdirs()
-  } else {
-    doRename()
-  }
+  fs.lstat(dest, function(err, statd) {
+
+    if ((statd) && (statd.isDirectory())) {
+      dest = path.join(dest, path.basename(source))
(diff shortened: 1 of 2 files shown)


jprichardson:
Thanks for your patience :) OK, I've reviewed this. In the `move` tests, I started switching from npm `testutil` to `'./lib/util'` which doesn't have any parameter as seen here: https://github.com/jprichardson/node-fs-extra/blob/master/test/lib/util.js#L19 so just so I'm clear, it's your intent that `TEST_DIR === EXISTING_DIR`, right?

jprichardson:
Ok, I have time to look at this again. Thanks for your patience!! So that I understand the nuances of this and what you're trying to achieve here, you want to emulate something like:

    mv /tmp/somedir/a-file /tmp/some-other-dir-that-already-exists/

So that now `/tmp/some-other-dir-that-already-exists/a-file` exists?

jprichardson:
I'm going to close this. Originally when I said that I was in favor, I misunderstood. Sorry about that. I'd like this feature in the future, but I want it to be explicit something like `fs.move('/some/file', '/dir/.')`. But `fs-extra` is a way off from this. Thanks for taking the time though, I really appreciate it.

PR #112, opened 2015-02-03
Copying a file into an existing directory

kosmasgiannis:
Here is the PR for the copy and copySync methods.

I hope it is OK.

### DIFF
--- a/lib/copy.js
+++ b/lib/copy.js
@@ -66,7 +66,15 @@ function copySync(src, dest, options) {
   options.filter = options.filter || function() { return true }
 
   var stats = options.recursive ? fs.lstatSync(src) : fs.statSync(src)
-  var destFolder = path.dirname(dest)
+
+  var statd = null
+  try {
+    statd = options.recursive ? fs.lstatSync(dest) : fs.statSync(dest)
+  } catch (e) {}
+  var destFolder = ((statd) && (statd.isDirectory())) ? dest : path.dirname(dest)
+  if ((statd) && (statd.isDirectory())) {
+    dest = path.join(dest, path.basename(src))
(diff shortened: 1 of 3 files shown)


jprichardson:
Yeah, so in these tests, `tesutil` is used from https://www.npmjs.com/package/testutil and not from here https://github.com/jprichardson/node-fs-extra/blob/master/test/lib/util.js#L19. So in theses tests, `DIR` will not be equal to `EXISTING_DIR`. Sorry for the confusion. I've been meaning to rid the entire library of `testutil` from npm. I'm just trying to make sure that I have your intent clear.

So to be 100% clear:
- In your `move()` PR:  `TESTING_DIR === EXISTING_DIR`
- In this PR: `DIR !== EXISTING_DIR`

kosmasgiannis:
Hi, 

in both cases I meant to use the same ``EXISTING_DIR`` which is different from ``TESTING_DIR`` and ``DIR``. That's why I call function ``createTestDir()`` with an argument.

It seems that I forgot to include in the PR the commit that modifies the createTestDir function which should now look like : 
```
function createTestDir(app) {
  if (! app) {
    app = 'fs-extra'
  }
  var dir = path.join(tmpdir(), 'test-' + app)
  if (fs.existsSync(dir)) {
    var files = fs.readdirSync(dir)
    files.forEach(function(file) {
      return rimraf.sync(path.join(dir, file))
    })
  } else {
    fs.mkdirSync(dir)
  }
  return dir
}
```

jprichardson:
Closing this, same reason as here: https://github.com/jprichardson/node-fs-extra/pull/111#issuecomment-91482871

PR #129, opened 2015-05-02
can you review this feature?

kimshinelove:
Can we developed a feature that reads the row and returns an array of additional features for file line review?


kimshinelove:
I will request to re-organize.

PR #130, opened 2015-05-02
Please review additional features.

kimshinelove:
It added a function that returns an array of rows read certain files.

Do we ask whether the review applies to the master branch.

In addition, the current master branch and the test failed in AppVeyor.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -82,7 +82,7 @@ fs.writeJSON = jsonFile.writeFile
 fs.writeJsonSync = jsonFile.writeFileSync
 fs.writeJSONSync = jsonFile.writeFileSync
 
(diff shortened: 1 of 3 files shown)


jprichardson:
1. The build fails on Windows. Probably should use `os.EOL` and join on an array for the test data. No need to fix it though.

2. Although, I definiltely see the utility here, I'll probably reject this as I'd like to keep this library simple and focus on getting the basic functionality working. I'll keep it open to think about it for awhile though.

Thanks for your interest.

jprichardson:
This is useful, but it's too early for `fs-extra`. There are many other priorities that I need to take care of before I'd accept this. Thanks for your interest.

Issue #101, opened 2015-01-04
fs-extra promisified


overlookmotel:
Just to let you know, I've published a module [fs-extra-promise](https://www.npmjs.com/package/fs-extra-promise) which promisifies fs-extra using bluebird.

For the most part you could just use `bluebird.promisify()`, but there are a few oddities which don't promisify properly. For example, `fs.exists()` doesn't use the usual node callback signature - it just returns `(result)` rather than `(err, result)` - so that breaks promisification.

jprichardson:
Cool, I'll add it to the readme! I like your addition of `isDirectory()`...
I'll probably add that :)

On Sun, Jan 4, 2015 at 11:08 AM, overlookmotel <notifications@github.com>
wrote:

> Just to let you know, I've published a module fs-extra-promise
> <https://www.npmjs.com/package/fs-extra-promise> which promisifies
> fs-extra using bluebird.
>
> For the most part you could just use bluebird.promisify(), but there are
> a few oddities which don't promisify properly. For example, fs.exists()
> doesn't use the usual node callback signature - it just returns (result)
> rather than (err, result) - so that breaks promisification.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jprichardson/node-fs-extra/issues/101>.
>



-- 
Simple & Secure Bitcoin Wallet: https://www.coinbolt.com
Bitcoin / JavaScript: http://cryptocoinjs.com
Follow me on Twitter: http://twitter.com/jprichardson

jprichardson:
https://github.com/jprichardson/node-fs-extra/commit/4729391768c20a6a39b67135694d8597bb9e628a#diff-04c6e90faac2675aa89e2176d2eec7d8

PR #132, opened 2015-05-06
isDirectory method

overlookmotel:
This PR adds convenience methods `isDirectory()` and `isDirectorySync()`.

They are shortcuts for `fs.stat()` followed by calling `.isDirectory()` on the result. These methods are in my [fs-extra-promise](https://github.com/overlookmotel/fs-extra-promise) module, but you said previously that you thought this could be useful to add to fs-extra, and it seems more natural to me to have them here. So here they are!

I've called them `isDirectory` to match the native node method name, but also aliased them as `isDir` to fit with the naming of other methods in fs-extra.

I didn't write any tests as (a) I'm not sure how to exactly within your testing framework and (b) the tests would basically be comprised of the same code as the methods themselves!

Please let me know if you'd like any amends to this PR.

### DIFF
--- a/lib/is-dir.js
+++ b/lib/is-dir.js
@@ -1,0 +1,18 @@
+var fs = require('fs')
+
+function isDirectory (path, callback) {
+  fs.stat(path, function (err, stats) {
+    if (err) return callback(err)
+
+    callback(null, stats.isDirectory())
+  })
+}
+
+function isDirectorySync (path) {
+  return fs.statSync(path).isDirectory()
+}
+
+module.exports = {
+  isDirectory: isDirectory,
+  isDirectorySync: isDirectorySync
+}
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ Methods
 - [ensureFileSync](#ensurefilefile-callback)
 - [ensureDir](#ensuredirdir-callback)
 - [ensureDirSync](#ensuredirdir-callback)
+- [isDirectory](#isdirectoryfile-callback)
+- [isDirectorySync](#isdirectoryfile-callback)
 - [mkdirs](#mkdirsdir-callback)
 - [mkdirsSync](#mkdirsdir-callback)
 - [move](#movesrc-dest-options-callback)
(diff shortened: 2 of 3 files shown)


jprichardson:
I can see the case for this, but I'm trying to keep this library as minimalistic as possible. I'm inclined to reject this. I'm going to leave it open to think it over for awhile. Thanks for your interest.

overlookmotel:
OK, no problem. I just made the PR as in previous conversation (https://github.com/jprichardson/node-fs-extra/issues/101#issuecomment-68645954) you said you liked it and would probably add it. So I thought I'd save you the trouble!

But I can see the argument in favour of minimalism, so it's totally up to you...

overlookmotel:
By the way, having looked through the way tests are constructed while doing the EEXISTS fix, if you do want to merge this, I can write some tests for it. Please let me know if you want this.

overlookmotel:
Just wondered if you'd had any further thoughts about whether you might like to include this functionality in fs-extra, or not?

jprichardson:
I have thought about it. I'm still undecided. But it's probably better to just make a decision now and move forward and revisit it later. So for now, I'll reject it. If in the future I change my mind, I'll ping you. Thanks for taking the time and the interest though :)

overlookmotel:
OK, no worries. I can understand your reasoning.

lib/is-dir.js
```diff
@@ -0,0 +1,18 @@
...
+function isDirectory (path, callback) {
+  fs.stat(path, function (err, stats) {
+    if (err) return callback(err)
+
+    callback(null, stats.isDirectory())
+  })
+}
+
+function isDirectorySync (path) {
+  return fs.statSync(path).isDirectory()
```
yvele: Isn't `isDirectorySync` supposed not to throw if the path doesn't exists? Because it will throw if you don't catch `statSync` for `ENOENT` errors.

The strange thing is that `isDirectory` swallows the errors and `isDirectorySync` not.. they don't have the same behavior.

@overlookmotel I think you have the exact same problem here: https://github.com/overlookmotel/fs-extra-promise

overlookmotel: @yvele I don't see it as a bug. I think it *should* throw if the path doesn't exist, same as calling `fs.stat()`/`fs.statSync()` would.

PR #133, opened 2015-05-07
Travis runs with `sudo: false`

overlookmotel:
Travis recently introduced a new container infrastructure which reduces queueing time for builds/test runs significantly.

This PR adds a line to the Travis config file to take advantage of this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ env:
   - TEST_SUITE=test
   - TEST_SUITE=coveralls
 script: "npm run-script $TEST_SUITE"
+sudo: false


overlookmotel:
Ah. One of the tests is failing, presumably to do with file permissions. I'll look into this and see if I can find a work-around.

overlookmotel:
Further commits I'm adding are to track down why the test is failing. I can't replicate it on my local machine (OS X) so having to push commits to run the tests on Travis. I'll let you know when I've resolved this (or given up!)

overlookmotel:
OK, I tracked down the bug. Looks like whatever version of Linux is running on the Travis container infrastructure throws an EEXIST error rather than ENOTEMPTY when you try to overwrite a directory while renaming.

Annoying that node doesn't conform these errors on different platforms to the same thrown error - it's meant to be cross-platform!

Anyway, it seems to work now and Travis gets builds underway much faster :)

jprichardson:
This is great! However, I don't like the the Travis modification and the `move` modification are made in the same PR. Would you be willing to split them up?

overlookmotel:
Have split it into two PRs. If you can merge the other one (https://github.com/jprichardson/node-fs-extra/pull/134), I'll rebase this on top of it so the tests pass.

overlookmotel:
Now rebased on top of v0.18.3, which included the EEXIST bug fix. So now all the tests pass and it's ready for merge.

jprichardson:
Excellent, thanks! (Dropping this here for reference: http://docs.travis-ci.com/user/workers/container-based-infrastructure/)

overlookmotel:
Thanks for the merge!

PR #134, opened 2015-05-08
Handle EEXIST error when clobbering dir

overlookmotel:
Some versions of Linux (including on the Travis container infrastructure) throw an EEXIST error rather than ENOTEMPTY when you try to overwrite a directory while renaming, and that wasn't being handled.

This PR fixes that.

### DIFF
--- a/lib/move.js
+++ b/lib/move.js
@@ -38,7 +38,7 @@ function mv (source, dest, options, callback) {
       fs.rename(source, dest, function (err) {
         if (!err) return callback()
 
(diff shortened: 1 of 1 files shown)


jprichardson:
Thanks for this! Btw, which Node.js version and Linux version did you notice this behavior on?

overlookmotel:
Thanks for merging. It was on Travis's container infrastructure. Reading Travis's output, seems that is Ubuntu 12.04.5 LTS.

PR #139, opened 2015-05-29
Preserve file timestamp on copy. Closes #138

### LINKED ISSUE #138: Preserve timestamps on copying files
On Unix it is possible to copy a file with the `-p` flag to preserve timestamps. It would be awesome if this option could be supported by fs-extra.

Basically I think that this should be quite easy to implement since Node.js already has the [`fs.utimes`](https://nodejs.org/api/fs.html#fs_fs_utimes_path_atime_mtime_callback) function that should have you covered.

There may be two options on how to use this:

- Either only on demand, i.e. if the user explicitly requests it,
- or implicitly, i.e. always when copying files.

What do you think about this?

goloroden:


jprichardson:
This isn't going to work. It needs to be modified in two ways:

1. It should async APIs.
2. There should be an included test.
3. What is the default behavior on Windows or Posix? There should be a flag added to control this behavior.

I really appreciate your interest and willingness to help. If you can't or don't want to make these modifications now, not a big deal. I will get to it eventually :)

goloroden:
Of course, you are right about this :-)

After a few hours of searching our actual problem, we did not think enough about how to do it properly. Sorry for that.

Unfortunately, to make it right, at least right now we do not have the time ... sorry :-(

jprichardson:
> Unfortunately, to make it right, at least right now we do not have the time ... sorry :-(

I figured, that's OK. I'll leave this PR open as a reminder for myself :)

goloroden:
This PR can then be closed, as it has been superseded by #141.

PR #153, opened 2015-07-19
Make preserveTimes also work on read-only files. Closes #152

### LINKED ISSUE #152: fs.copy fails for read-only files with preserveTimestamp=true
When trying to `fs.copy` a directory with `preserveTimestamps` set to `true`, most of the times everything works fine.

The situation where it does _not_ work is when you try to copy a directory that contains files that the user does not have write access to, e.g. a `.git` directory (some of the files below `.git/objects/pack` are read-only for the user).

Then, `fs.copy` fails with an `EACCES` error.

Basically, I see two options on how to deal with this:
- You could ignore that setting the timestamp does not work.
- You could enable write access, set the timestamp, and then remove write access again.

While the first one would not _always_ be correct, the second one potentially produces new problems: It would only work if the file is owned by the user, i.e. if the user has the rights to change file permissions.

What do you think about simply ignoring these errors? Maybe this could be coupled with a `force` option to enforce copying files, even if some of the timestamps could not be set.

PS: Could it be an option to change this call to [`fs.futimes`](https://github.com/jprichardson/node-fs-extra/blob/146dfecd554cf49af0fbd1d6a223662bcc01763c/lib/util/utimes.js#L55-L61) to a call to `fs.utimes`? AFAICS this would solve the issue.

PPS: I have tried changing [file mode](https://github.com/jprichardson/node-fs-extra/blob/146dfecd554cf49af0fbd1d6a223662bcc01763c/lib/util/utimes.js#L55) from `r+` to `r`. Apparently it still works, although you don't have write access to the file. At least your tests still run through ;-). Maybe this is the easiest way to solve the issue?

goloroden:


jprichardson:
Thanks. I don't like how there isn't any tests though. I'll have to dig into the Windows failure a bit more...

PR #161, opened 2015-08-04
ensure symlink & link

reggi:


reggi:
What I'm getting from this is: 

1. `isAbsolute` isn't in older version of node, there's a fill [sindresorhus/path-is-absolute](https://github.com/sindresorhus/path-is-absolute)
2. Some of my tests use paths that are posix specific. and expect urls in a specific format.

jprichardson:
Fantastic work! It looks like you put a lot of time into this, thank you! 

1. I'm ok with including the polyfill `path-is-absolute` until we drop Node 0.10 support.
2. Why the inclusion of `chai`? Why not just use basic `assert`?
3. Would it be to much work to fix the POSIX specific paths? So that the tests pass on Windows?

PR #162, opened 2015-08-04
ensure symlink & link

reggi:


reggi:
@jprichardson had no clue that the checks would reset when I forced a commit. ¯\_(ツ)_/¯

reggi:
I tried moving all the `expect` calls over to the native `assert` calls and I ran into a snag with `expect().to.contain`, don't know a way of doing that without `chai`.

reggi:
@jprichardson Yay! It passed! Wooooohoo. Thoughts?

jprichardson:
Awesome!

Regarding `chai`, why not...

Instead of this:

```js
expect(fs.readdirSync(dstDir)).to.contain(dstBasename)
```

```js
assert(fs.readdirSync(dstDir).indexOf(dstBasename) >= 0)
```

reggi:
@jprichardson updated & passed removed chai as a dev-dependancy.

reggi:
@jprichardson Just realized this still creates the directories even if the link throws an error. Thoughts? Should I check if both files exist then `mkdirs`? If they don't exist what errors should be returned? I could use `lstat` instead of `exist` and run search and replace for the function name `lstat` in the native node error. 

So for both `srcpath` and `dstpath` with lstat:

```
"Error: ENOENT: no such file or directory, lstat './mysource.txt'".replace("lstat", "link")
```

This would keep errors consistent across different versions of node.

jprichardson:
> Just realized this still creates the directories even if the ensureLink throws an error. 

Good catch.

> Should I check if both files exist then mkdirs? If they don't exist what errors should be returned? I could use lstat instead of exist and run search and replace for the function name lstat in the native node error message.

I really like this idea.

reggi:
moving branch off of master to a squashed version of the feature development branch

Issue #164, opened 2015-08-04
Feature Request: ensureFile to take optional argument for file content


reggi:
What are thoughts on `ensureFile` also being able to take the string argument for file content. So if I run this:

```
fse.ensureFileSync("./tmp/paths/do/not/exist/readme.md", "#document)
```

The file `readme.md` is created with ensured dirs `/tmp/paths/do/not/exist/` with the content `#document`.

reggi:
It's only a couple of lines to add this functionality. Keeping this here for safe keeping:

```javascript
var path = require('path')
var fs = require('graceful-fs')
var mkdir = require('../mkdirs')

function createFile (file, content, callback) {
  callback = (typeof content === 'function') ? content : callback
  content = (typeof content === 'function') ? '' : content

  function makeFile () {
    fs.writeFile(file, content, function (err) {
      if (err) return callback(err)
      callback()
    })
  }

  fs.exists(file, function (fileExists) {
    if (fileExists) return callback()
    var dir = path.dirname(file)
    fs.exists(dir, function (dirExists) {
      if (dirExists) return makeFile()
      mkdir.mkdirs(dir, function (err) {
        if (err) return callback(err)
        makeFile()
      })
    })
  })
}

function createFileSync (file, content) {
  content = (typeof content === 'function') ? '' : content

  if (fs.existsSync(file)) return

  var dir = path.dirname(file)
  if (!fs.existsSync(dir)) {
    mkdir.mkdirsSync(dir)
  }

  fs.writeFileSync(file, content)
}

module.exports = {
  createFile: createFile,
  createFileSync: createFileSync,
  // alias
  ensureFile: createFile,
  ensureFileSync: createFileSync
}
```

jprichardson:
I'd support it. But I'd like the second parameter to be an object and the object parameter `contents` would be used to denote the contents.

Behavior:

1. If `contents` is not specified, the method just verifies that the file exists. If not, it creates it.
2. If `contents` is specified, it verifies that the file exists, and overwrites the existing content with `contents`. If file does not exist, just create file with new contents.
3. `contents` should allow either `string` or `buffer`.

Thoughts?

reggi:
This is great feedback @jprichardson thanks! Here are some of my thoughts.

The current behavior of ensure should "ensure that the file exists" and do all it can to abide by the `If the file, dir, symlink, link already exists, it is NOT MODIFIED.` motto.

[...]

PR #165, opened 2015-08-06
ensure for link & symlink

reggi:


reggi:
@jprichardson

Added:

* error messages use lstat and replace the string `lstat` with `ensureSymlink` and `ensureLink` respectivley
* will not create recursive destination directories on error 
* branched out feature and dev so you can see diffs between pull-requests [here](https://github.com/reggi/node-fs-extra/commits/dev/ensurelink-ensuresymlink)

reggi:
Should an error be thrown if the destination path exists?

jprichardson:
> Should an error be thrown if the destination path exists?

I don't think so.

Thanks for your hard work on this, I really appreciate it! Published `0.23.0`.

reggi:
The functions currently return an error for async and throw for sync if the destination path exists because that's the native functionality of `fs.link` and `fs.symlink` is exposed.

jprichardson:
Yeah, I think that they should be normalized to not throw errors if the destination exists. I mean, the intent of the caller of these methods is to just make sure that they exist regardless of whether they existed before said method was called. Don't you think?

reggi:
As expressed [here](https://github.com/jprichardson/node-fs-extra/issues/164#issuecomment-128069620) I feel weird throwing the err away. My current usage for this function is as follows, please forgive the promises.

```
fse.ensureLinkAsync(pkg.main, paths.mainDst)
.then(debugThen('linking main %s -> %s', pkg.main, paths.mainDst)).catch(debugCatch),
```

So for control-flow it's useful to know if something has been created or not. 

Thats why [I expressed the need for `ensureWarn` functions](https://github.com/jprichardson/node-fs-extra/issues/164#issuecomment-128083607). 

I agree though with this:

> the intent of the caller of these methods is to just make sure that they exist regardless

Crafting PR(s) now.

reggi:
https://github.com/jprichardson/node-fs-extra/pull/169

PR #167, opened 2015-08-06
Adds basic (string, buffer) support for ensureFile content

reggi:


jprichardson:
I really appreciate your initiative here. I do wish that you would have waited though so that we could elaborate on our discussion in https://github.com/jprichardson/node-fs-extra/issues/164. I'm against this PR, but will leave it open for a bit to think about it.

reggi:
@jprichardson I appreciate you leaving this open to consider it, I'd love to chat more about it. Personally I don't think that offering this out of the box is a bad idea, we can always come back to extending the arg functionality to be more complex later.

jprichardson:
Closing this for now as I really want to get this module to `1.0` and I'm still undecided on how to best handle this. Tagged with `future` though so that I can revisit this feature. I do really appreciate your previous PRs and help :)

PR #169, opened 2015-08-06
suppress ensureLink & ensureSymlink dest exists error

reggi:
Here's the dest error suppressing for ensureLink thoughts? I'd do the same for ensureSymlink

### DIFF
--- a/lib/ensure/__tests__/link.test.js
+++ b/lib/ensure/__tests__/link.test.js
@@ -34,7 +34,7 @@ describe('fse-ensure-link', function () {
     [['../foo.txt', './link.txt'], 'file-error', 'file-error'],
     [['../dir-foo/foo.txt', './link.txt'], 'file-error', 'file-error'],
     // error is thrown if destination path exists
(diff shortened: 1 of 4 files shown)


reggi:
Added symlink in here too, and rebased to one commit.

jprichardson:
Thanks!

jprichardson:
Hmm... what if the link exists but it's to another source?

reggi:
@jprichardson This doesn't check the type of the file, or if the content is the right content. The way I understood ensure is, "ensure it's on the system" if it exist's in any way shape or form it's ensured, no? This could check if it's ensured as a specific type of path (using stat) and if the content is what it should be. But then what? Throw an error? I'm pretty comfortable with this as it is. I do want to see the [extensible range of ensure derivative functions that I talk about here](https://github.com/jprichardson/node-fs-extra/issues/164#issuecomment-128082003) I'd chalk all this functionality into a `ensureMatch` derivative.

> tl;dr I'm ok with if it's to another source, it's ensured to exist

jprichardson:
Fair points. We'll leave it as is for now.

PR #170, opened 2015-08-07
More robust handling of errors moving across virtual drives

nicks:
Discovered while investigating
https://github.com/Medium/phantomjs/issues/341
where moving a directory across virtual drives caused an
EXDEV error that shunted things down the "move file"
instead of "move directory" codepath


jprichardson:
Thank you!

PR #189, opened 2015-10-31
move.test fix to support CRLF on Windows

### LINKED ISSUE #188: move.test failing on windows platform
There are 7 tests which fail on windows but work on Linux.
This is because when git fetches the test data files it converts the line endings from \n to \r\n (since core.autocrlf is typically true with git on windows).

This fails on Windows but succeeds elsewhere:
```
        assert.ifError(err)
        assert.strictEqual(contents, 'sonic the hedgehog\n')
```
I believe the best way to fix this is to use regex:
```
        var expected = /^sonic the hedgehog\r?\n$/
        assert.ifError(err)
        assert.ok(contents.match(expected), contents + ' match ' + expected)
```

bartland:
move.test modified to support testing files with both LF (NIX) and CRLF (Windows) line endings. This commit fixes #188.

### DIFF
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -60,8 +60,9 @@ describe('move', function () {
     fse.move(src, dest, function (err) {
       assert.ifError(err)
       fs.readFile(dest, 'utf8', function (err, contents) {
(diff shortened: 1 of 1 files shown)


jprichardson:
Excellent, thanks :+1:

PR #190, opened 2015-10-31
copySync to overwrite destination file if readonly and clobber true

### LINKED ISSUE #183: with clobber true, copy and copySync behave differently if destination file is read only
copy succeeds because it will first do an fs.unlink() on the destination file (in ncp.js).
copySync fails. 

To align the behaviour, in copy-file-sync.js change:

    if (fs.existsSync(destFile) && !clobber) {
      throw Error('EEXIST');
    }

to

    if (fs.existsSync(destFile)) {
      if (clobber) {
        fs.unlinkSync(destFile);
      } else {
        throw Error('EEXIST');
      }
    }

bartland:
copySync updated to overwrite destination file if readonly and clobber set to true. Added test case. This commit fixes #183.

### DIFF
--- a/lib/copy-sync/__tests__/copy-sync-file.test.js
+++ b/lib/copy-sync/__tests__/copy-sync-file.test.js
@@ -183,6 +183,20 @@ describe('+ copySync()', function () {
             assert.strictEqual(destData, destDataNew)
           })
         })
+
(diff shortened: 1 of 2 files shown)


jprichardson:
Thank you for your help! Published in 0.26.1

PR #194, opened 2015-11-19
allow ensureFile(Sync) to provide data to be written to created file

tonylukasavage:
> This PR includes tests and docs for a backwards compatible change to the API.

Essentially I've changed the signature of `ensureFile(file, callback)` to `ensureFile(file, [data], callback)`, adding an optional `data` parameter that can be used to populate a created file. I use fs-extra frequently and find myself needing this functionality fairly often. For example, if I want to make sure an JSON configuration file exists, and if not put bare minimum valid JSON in it, I might use this new version of the API like this:

```js
fse.ensureFile('file.json', '{}', function(err) {
    // do stuff...
});
```

The change is modeled along the same lines as node.js's `fs.writeFile()` API. At this time it does not _complete_ that API, meaning this PR does not attempt to additionally add the optional `options` parameter that is available in `fs.writeFile()`. Having said that, if adding that aspect would make this PR more acceptable, I'd be happy to implement that as well, making `ensureFile()` and near mirror of the `fs.writeFile()` API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,9 @@ fs.emptyDir('/tmp/some/dir', function (err) {
 ```
 
 
-### ensureFile(file, callback)
+### ensureFile(file, [data], callback)
 
-Ensures that the file exists. If the file that is requested to be created is in directories that do not exist, these directories are created. If the file already exists, it is **NOT MODIFIED**.
+Ensures that the file exists. If `data` is provided, it will be written to the created file, otherwise an empty file will be written. If the file that is requested to be created is in directories that do not exist, these directories are created. If the file already exists, it is **NOT MODIFIED**.
 
 Alias: `createFile()`
 
--- a/lib/ensure/__tests__/ensure.test.js
+++ b/lib/ensure/__tests__/ensure.test.js
@@ -45,6 +45,20 @@ describe('fs-extra', function () {
         })
       })
     })
+
+    describe('> when file does not exist and data is provided', function () {
+      it('should create the file with data', function (done) {
+        var file = path.join(TEST_DIR, 'dir/that/does/not/exist', 'file.txt')
+
+        assert(!fs.existsSync(file))
+        fse.ensureFile(file, 'test', function (err) {
+          assert.ifError(err)
+          assert(fs.existsSync(file))
+          assert.equal(fs.readFileSync(file, 'utf8'), 'test')
+          done()
+        })
+      })
+    })
   })
 
   describe('+ ensureFileSync()', function () {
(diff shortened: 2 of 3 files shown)


tonylukasavage:
Looks like the appveyor setup has an issue with the `nodejs=2.3` configuration... all other tests are passing and it doesn't appear to be related to this PR directly.

jprichardson:
Why not use `outputFile`?

tonylukasavage:
I could, but I'd still have to do an `exists` check before executing `outputFile`, since i only want to write the file if it doesn't exist. I do this frequently for tooling projects that require an initial configuration.

tonylukasavage:
Any additional thoughts here given the subtle but important distinction between `outputFile` and this new functionality added to `ensureFile`?

jprichardson:
> I could, but I'd still have to do an exists check before executing outputFile, since i only want to write the file if it doesn't exist. I do this frequently for tooling projects that require an initial configuration.

If I were to see `ensureFile('some data')` I'd think that the call to that method will always make sure that the file exists and the data is present as passed into the method.

Detailed discussion and reference: https://github.com/jprichardson/node-fs-extra/issues/164

tonylukasavage:
OK, I've read through the content of #164 and I get the context and concerns, though I'm not sure where that discussion leaves this functionality. That is, I'm not clear what next steps would be to include it in the API. Based on the depth of the discussion I fear this particular "ensure, but add data if it doesn't exist" functionality is languishing in limbo.

Based on the conversation n #164 and what's been proposed here, what would you like to see happen in the API to support this functionality? Even if it's well beyond the scope of this relatively simple change, since it seems the prior conversation mentions parallels with other fs-node-extra APIs beyond just ensureFile(). I think I'm missing your vision of what else this might impact and how it will be perceived by users of the module, and having that vision will likely help me put together a better, more complete PR.

tl;dr: I would find this functionality really useful and would love to get it merged. Lemme know what I gotta do, even if it's a lot more work, to get it in.

nicosommi:
+1 to get this merged

nicosommi:
BTW, Currently this is how it looks like
stat(mypath)
  .catch(() => outputJson(mypath, initialData))

jprichardson:
@tonylukasavage I'm sorry I didn't respond right away... 😞  

> If the file does not exists, initialize it with the data, but if it exists, return the data with no modifications.
Maybe the second parameter should be named [initialData]

@nicosommi I'm not sure that would fit under a method name `ensureFile`.

Overall, at this time, I think that `fs-extra` needs to be as slim as possible until stability (1.0.0) is achieved. Even at the expense of some people using it. Once methods are added, it's very hard to remove them and I think what's being requested here is simple function - you could create a small little utility method and call it a day.

dmatora:
@tonylukasavage @jprichardson how about creating ensureJSON?
We do have both readFileSync and readJSONSync, doesn't make much sense why we have ensureFile, but not ensureJSON

RyanZim:
@dmatora what would `ensureJSON`'s behavior be?

Issue #193, opened 2015-11-18
`fs.copy` fails silently if source file is /dev/null


raybellis:
Using bluebird with `fs.copyAsync`, my program finished with no error despite a `.catch(console.error)` when I tried to copy `/dev/null` onto a file.

jprichardson:
Do you get the same problem with just `fs.copy`?

raybellis:
Let me try that and see...

raybellis:
It appears to fail, silently, without creating the file:

    var fs = require('fs-extra');

    fs.copy('/dev/null', '/tmp/foo', function(err) {
        console.log(err);
    });

jprichardson:
Thanks. Verified. Just hangs on OS X 10.11 / Node 4.2.1. 

Surely this is the offending line: https://github.com/jprichardson/node-fs-extra/blob/0.26.2/lib/copy/copy.js#L15

Ideas on a practical solution? I'd hate as something could be symlinked to /dev/null right? Surely others have experienced this... a cursory Google search didn't return much.

Thanks for your help.

jprichardson:
Nope, this is the offending line:

https://github.com/jprichardson/node-fs-extra/blob/0.26.2/lib/copy/copy.js#L24

Is there something about the `stats` of `/dev/null` that I could use to reliably detect it?

raybellis:
No, it's not there, either - it's the `getStats()` function in `lib/copy/ncp.js` which doesn't have a branch to take if the given file isn't either a directory, a plain file, or a symlink.

raybellis:
I think the check in that function for `stats.isFile()` needs the addition of:

    || stats.isBlockDevice() || stats.isCharacterDevice()

jprichardson:
I would love a pull request :) I have a test here: https://github.com/jprichardson/node-fs-extra/commit/631ed20f7bd8a817346073efa59f93f19c77da48

jprichardson:
Updated test. Thanks again for your help, it's much appreciated!

PR #204, opened 2015-12-17
allow copy of block and character device files

raybellis:
allow copying of character or block devices (NB: copies contents, not the inode)


jprichardson:
Thank you! (For record: references #193)

PR #215, opened 2016-01-27
fse.copy throws error when only src and dest provided

ioncreature:
I have a lot of legacy code which uses old version(0.12.0) of fs-extra.
Old code does not work with new fs-extra due to it uses fse.copy with only 2 arguments `src` and `dest`.
Let's improve backward compatibility, guys

### DIFF
--- a/lib/copy/copy.js
+++ b/lib/copy/copy.js
@@ -11,6 +11,7 @@ function copy (src, dest, options, callback) {
     options = {filter: options}
   }
   callback = callback || function () {}
+  options = options || {}
 
   fs.lstat(src, function (err, stats) {
     if (err) return callback(err)


jprichardson:
Thank you, this is a documentation failure on my part as `fse.copy()` should have never have been used with two arguments OR I should have anticipate that people would have used it in such a way and added a test to ensure that it doesn't fail.

jprichardson:
Published as `0.26.5`.

ioncreature:
@jprichardson Thanks for quick response!
Could you update it on npm?

It would be nice to use this patch right now.

Thanks again.

ioncreature:
Wow, you are reading my thoughts )

Issue #70, opened 2014-07-09
Not filter each file, stops on first false


ayauka:
Files only filter each file while ever returns true, if the first file on the folder or the folder name dont pass the filter, the copy stops.

jprichardson:
Closing. Please try the latest version. If it does not work, please create a new issue with a snippet. Thank you.

PR #216, opened 2016-01-29
fix #187 #70 options.filter bug

### LINKED ISSUE #187: Not filter each file, stops on first false
I've got same issue as #70 .

Im using latest release. 

"fs-extra": "^0.26.0",

eos3tion:
#187 #70
Files only filter each file while ever returns true, if the first file on the folder or the folder name dont pass the filter, the copy stops.


jprichardson:
Thanks for much for this!

1. Would you run `npm test` and clean up the errors?
2. Also, would you add a test case? The test should fail before your patch and your patch should make it pass?

Thanks again!

PR #220, opened 2016-02-07
Non-breaking feature: add option 'passStats' to copy methods.

shennan:
I've found a few use cases where I've wanted to query the file/directory statistics from within the `filter` function handler. Something like this:

``` js
copy('some/path', 'other/path', {
  filter: function (p, stats) {
    return stats.isDirectory();
  }
});
```

All sorts of useful information could be garnered from the `stats` object that could inform the decision to skip the path in the `filter` callback. Seeing as the stats are being fetched anyway, it seemed like a pretty straight-forward adjustment for a lot of extra gain.

**_Why make it an option? Why not just pass it?_**

Because there could be an argument for greater speed in not using it. The `fs.stat` is always called, but in the asynchronous `copy`, the stats are only fetched after the path has been filtered. In the case of passing the `{ passStats: true }`, we can signal to the copy method that we'd like to delay the filter call until further down the chain of checks, so that we can make use of the inevitable call to `fs.stat`.

On top of the code, I have also altered the docs to reflect the new feature.

Let me know your thoughts.


shennan:
Let me know if this is something you'd be willing to pull in and I'll make sure the tests are passing.

jprichardson:
I really appreciate you making the case here. I prefer to err on the side of simplicity. You've definitely made a good case on the need for the option given the performance hit of doing the actual reading. I like to keep the options consistent with the `sync` versions though, so the option would have to be added there. This adds more complexity.

What's your use case?

What I'd really like to do is rewrite `copy()` using `walk()`. `walk()` still needs to include a read directory filter, but you could tap into the stream and filter the stats appropriately there. What are your thoughts about that?

shennan:
I agree that simplicity is something to strive for. The key is hitting the sweet spot between simplicity and utility. For me, including stats in the filter callback is extremely useful in making real decisions about whether to include the file in the copies. The `fs.Stats` Class is also an established node class, so many will be able to make use of the feature without needing to learn anything new in the api.

I guess, the argument between simplicity and utility here is the specific one of whether to enable the feature through an options flag (`passStats`) or to just pass them anyway. I'm in favour of taking the very small performance hit and just passing it, so as not to pollute the configuration of the method. The reasons:

- It is simpler (as you have expressed a wish for)
- The performance hit is minimal because...
- All files that pass the filter in `copy` will inevitably be requesting the stats object further down in the code.
- As it stands, `copySync` requests the stats object before invoking `filter`, and nobody has complained about that as of yet. So `copySync` will see no change in performance.
- Any calls to `copy` that don't include a filter have no performance hit whatsoever.

So my vote would be to drop the `passStats` flag and just pass the stats over to the filter method as the second argument, for all `filter` handlers.

:-)

jprichardson:
All of your use cases seem like an excellent reason to implement `copy()` on `walk()` and make it a stream. I'd like your thoughts on my comment above:

> What I'd really like to do is rewrite copy() using walk(). walk() still needs to include a read directory filter, but you could tap into the stream and filter the stats appropriately there. What are your thoughts about that?

shennan:
Apologies, I was a little sleepy when I read your message so I missed the comment about aggregating `copy` and `walk`.

I think it's a good idea, but I'm unsure of how you would do it without breaking changes. Could you give me an example of the syntax you would use for a `walk`-ified `copy`?

Would you be dumping the current copy methods etc? And would it degrade gracefully for 0.10 & 0.12?

shennan:
@jprichardson Still eager to hear your thoughts/plans about copy/walk. A syntax example of your idea would be good.

jprichardson:
I think this could be done while maintaining compatibility. Essentially if a callback is passed as the last parameter, that'd be the same as the call to `end` on the stream. `fs.copy` itself would return a readable stream. Each `data` event would be the item to be copied. Filtering would present a bit of a challenge that I'll have to think more about.

shennan:
@jprichardson I suppose that's what I'm driving at; how are you imagining filtering would be achieved in this instance?

jprichardson:
@shennan I'm not sure of the best way to handle it at the moment. I think you'd probably just keep the same filtering mechanism that exists now and convert that to `walk()` filters - which I need to change a bit.

develar:
stat is required, 👍 https://github.com/electron-userland/electron-builder/issues/701 Without `stat.isDirectory` it is not possible to decide — should match be partial or not (minimatch pattern).

develar:
@shennan What's left to finish this PR?

develar:
Ah, I see — `running` is not decremented and copy hangs.

develar:
https://github.com/develar/node-fs-extra/commit/779c75c0106235ff12afeedd2cb1a5b39ac2d2d9

jprichardson:
It seems to me that it should just get passed and there should be no `passStats` option. I'm sure this could work... just needs tests and implemented in `copySync` as well. Although, I'm considering splitting out the sync methods, but I think it's too premature to do that as I'd like `async/await` in core first.

develar:
@jprichardson Thanks for quick answer, because I don't want to maintain my own fork :) Will try to finish this PR next week.

jprichardson:
@develar sure. I forgot about @shennan's original point:

> Why make it an option? Why not just pass it?
>
> Because there could be an argument for greater speed in not using it. The fs.stat is always called, but in the asynchronous copy, the stats are only fetched after the path has been filtered. In the case of passing the { passStats: true }, we can signal to the copy method that we'd like to delay the filter call until further down the chain of checks, so that we can make use of the inevitable call to fs.stat.

So on a large test data set, it should be quantified how much slower is it to just always pass the stats.

develar:
> it should be quantified how much slower is it to just always pass the stats.

I think, we should not decide for user. I have old machine with slow HDD to perform test, but I think that this functionality is not required for all users (you see — how many votes here ;)?) and if we can avoid stat — we should avoid.

shennan:
Yes I agree with @develar 

I can fix the failing CI tests on my original PR this coming week and we can sign this off soon.

How do you feel about that @jprichardson? It's a non-breaking feature at the end of the day.

shennan:
Lint and unit tests are now passing.

@jprichardson if you're still not convinced by the PR, let me know how you want to proceed.

jprichardson:
@shennan thank you for stepping up and doing this. Please do a benchmark on the performance difference of just passing the stats vs not.

RyanZim:
@shennan Ping?

shennan:
Hiya,

Message about benchmarking seemed to pass me by.

Is there a specific benchmarking package I should use? Some more specifics on input/output would be good. Thanks.

jprichardson:
> Is there a specific benchmarking package I should use? Some more specifics on input/output would be good.

Good question. I have a lot of respect for this dev, so you could look at this: https://github.com/indutny/elliptic/tree/master/benchmarks

I really just want you to quantify your original statement:

> Because there could be an argument for greater speed in not using it.

RyanZim:
Those tests use https://github.com/bestiejs/benchmark.js

shennan:
[Here](https://gist.github.com/shennan/f93af25bf032981c714d17d37a7462b1) is a Gist of the benchmark code I've written. Note that it requires an npm link to the local repository where my `passStats` changes have been made. For use in the wild, you will want to grab this pull request and link using the [instructions here](https://docs.npmjs.com/cli/link).

The output is as follows:

```
Asynchronous
WithFilter x 387 ops/sec ±12.54% (44 runs sampled)
WithFilterAndStats x 370 ops/sec ±16.96% (41 runs sampled)
Fastest is WithFilter,WithFilterAndStats
Synchronous
WithFilter x 811 ops/sec ±8.89% (69 runs sampled)
WithFilterAndStats x 852 ops/sec ±18.51% (70 runs sampled)
Fastest is WithFilter,WithFilterAndStats
```

And another try:

```
Asynchronous
WithFilterAndStats x 436 ops/sec ±13.80% (45 runs sampled)
WithFilter x 515 ops/sec ±13.94% (59 runs sampled)
Fastest is WithFilter,WithFilterAndStats
Synchronous
WithFilterAndStats x 523 ops/sec ±14.28% (45 runs sampled)
WithFilter x 718 ops/sec ±12.13% (60 runs sampled)
Fastest is WithFilter
```

This test was done with 5 files to copy from src to dest, of about 92kb each. Perhaps I should try with a bigger data set to give more conclusive results? Benchmarking is not my forte...

RyanZim:
This is merge-conflicted beyond repair, and we now allow `async` filter functions, making getting stats much easier. Closing.

PR #221, opened 2016-02-12
Update README.md

Cellule:
Fixes links for `outputFile()` and `outputJson()` in the table of content.
List `options` in a bullet point list. The current format is hard to read as it doesn't put different options on a newline.
Removed `copySync()` and added `Sync` description to `copy()` to be more concise with the rest of the doc.
Added a section to explicitly show the difference between `Sync` and `Async` idioms.

[Review on Reviewable](https://reviewable.io/reviews/jprichardson/node-fs-extra/221)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,11 +61,35 @@ var fs = require('fs')
 var fse = require('fs-extra')
 ```
 
+Sync vs Async
+-------------
+Most methods are async by default (they take a callback with an `Error` as first argument).
+
+Sync methods on the other hand will throw if an error occurs.
+
+Example:
+
+```js
+var fs = require('fs-extra')
+
+fs.copy('/tmp/myfile', '/tmp/mynewfile', function (err) {
+  if (err) return console.error(err)
+  console.log("success!")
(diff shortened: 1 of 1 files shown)


jprichardson:
Thank you, these changes are very good. One concern, I was in the process of separating out all of the `sync()` methods into their own section because new users were getting confused about whether they should have a callback or not.

What are your thoughts? Do you think each `sync` method deserves its own section or do you think that the `sync vs async` is adequate?

Thanks.

Cellule:
I find it hard to judge on this point because I am now used to both idioms and they come naturally. 
I added that section to help those that have issues understanding the difference. I think it's a good step toward clarity, but possibly not enough.

Looking at node [file system doc](https://nodejs.org/api/fs.html#fs_fs_readfile_file_options_callback), the sync method simply refers to the async version and gives details about the return value.
Maybe following in their footstep is a good way to make everyone happy.

jprichardson:
Good points. Thanks for the feedback.

PR #222, opened 2016-02-17
A clearer description of emtpyDir function

yetanotherusernamebecausegithubisbugged:
I was confused by the description so I propose to commit a clearer version.


jprichardson:
Thank you!

PR #228, opened 2016-03-14
Throw error when target is empty so file is not accidentally zeroed out

glortho:
This happens if you set source and destination to the same file. Note that `fs.rename` handles this gracefully, and users may expect the same with fs-extra's `copy`. This PR is probably not the best way to handle this, and does not address all cases, but I hope this will open discussion on this issue, since in any event a user copying a file to self would not want or expect the file to remain, but now at length=0.


jprichardson:
Yea, this would be a good bug to fix. Thanks. What about checking if they're the same at the start of the `copy()` method?

PR #229, opened 2016-03-15
fix 'TypeError: callback is not a function' in emptyDir

daysv:


jprichardson:
Thanks for your help!

Issue #191, opened 2015-11-03
How to copy symlinks to target as normal folders


screendriver:
Can I set the `-L` option like in the unix command `cp`?

From the man page:

> ‘-L’ ,‘--dereference’ - Follow symbolic links when copying from them. With this option, cp
> cannot create a symbolic link. For example, a symlink (to regular file) in the source tree will be
> copied to a regular file in the destination tree. 

See also here at [superuser](http://superuser.com/questions/216919/how-to-copy-symlinks-to-target-as-normal-folders)

jprichardson:
Try setting `deference` to `true`:

```js
fs.copy(src, dest, { dereference: true }, callback)
```

Let me know if that works. I'm gonna leave this open, as that should be documented.

screendriver:
Thank you, it's working :+1:

jhnns:
Thx @jprichardson. Just being curious: why is this option not listed in the README?

jprichardson:
@jhnns it's because it doesn't exist in `copySync`. It should be implemented in `copySync` as well.

jhnns:
Sounds reasonable :+1:

jprichardson:
Implemented in https://github.com/jprichardson/node-fs-extra/pull/235

PR #235, opened 2016-04-15
Adds symlink dereference option to `fse.copySync` (#191)

rjz:
Carries forward the behavior and specs governing the existing (async-only) implementation and updates the documentation.

Thanks for considering this change!


jprichardson:
Fantastic, thank you - well done on the PR (commit structure, tests, docs) as well!

PR #260, opened 2016-07-25
use more specific path for method require

hefangshi:
We encounter a problem when we upgrade fs-extras from 0.18.3 to 0.30.0, all extra method like fs.move is undefined after we upgrade our service.

Our deploy system is a increment deploy system, so every history file was  on the machine.

In 0.18.3 the file path is `lib/move.js` and in 0.30.0 is `lib/move/index.js`, they can both by required by using `require('./move')`, so when we upgrade the fs-extras lib, the new `index.js` will require('lib/move.js') rather than `lib/move/index.js` since old file is still exist.

I think it won't hurt anything if we change require path to a more specific path, and will get a better compatibility.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,16 +10,16 @@ Object.keys(gfs).forEach(function (key) {
 
 var fs = fse
 
-assign(fs, require('./copy'))
-assign(fs, require('./copy-sync'))
-assign(fs, require('./mkdirs'))
-assign(fs, require('./remove'))
-assign(fs, require('./json'))
-assign(fs, require('./move'))
-assign(fs, require('./empty'))
-assign(fs, require('./ensure'))
-assign(fs, require('./output'))
-assign(fs, require('./walk'))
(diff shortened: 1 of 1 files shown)


jprichardson:
I don't see how this really changes anything? The `require` algorithm will look for `index.js` if it finds a `directory`.

hefangshi:
Yes you are right, and the code is working great at in normal scenes.

The thing is our deploy system is a increment deploy system, so there will be `lib/move.js` and `lib/move/index.js` at the **same time** after we upgrade fs-extras from 0.18.3 to 0.30.0, and the `require` algorithm will find old `lib/move.js` be default, and the code was broken.

RyanZim:
@hefangshi This is a problem with your deploy system, not fs-extra IMO. Going to close this for the time being.

hefangshi:
It has no harm for fs-extra, so why don't accept it?

PR #268, opened 2016-08-24
port fix for mkdirp issue #111

pabigot:
The original mkdirp mis-handles the fourth parameter when a directory
component is missing.  The same issue is present in the fs-extra
variant.

See: https://github.com/substack/node-mkdirp/issues/111

### DIFF
--- a/lib/mkdirs/mkdirs.js
+++ b/lib/mkdirs/mkdirs.js
@@ -40,7 +40,7 @@ function mkdirs (p, opts, callback, made) {
         mkdirs(path.dirname(p), opts, function (er, made) {
           if (er) callback(er, made)
           else mkdirs(p, opts, callback, made)
-        })
+        }, made)
         break
 
       // In the case of any other error, just see if there's a dir


jprichardson:
Great, thanks. Please add a test.

pabigot:
@jprichardson I'll probably do that but I'm going to wait until upstream decides what to do, since IMO variant behavior between fs-extras and mkdirp would be worse than both being wrong the same way.

manidlou:
@jprichardson @RyanZim what is this one's destiny?!

RyanZim:
Upstream mkdirp has never merged the patch, until then, I'm not merging it either.

manidlou:
> Upstream mkdirp has never merged the patch, until then, I'm not merging it either.

gotcha!

pabigot:
FWIW, since upstream hasn't seen any updates since before this bug was reported I expect it to remain unmerged.  At least it's logged in both places, and a fix is available for whoever really needs it.

RyanZim:
Ah, wasn't aware of that. At fs-extra, we've never documented the `made` parameter AFAIK; I'm thinking we should just remove this undocumented feature in the next major release, especially since Node.js will be adding a `recursive` flag to `fs.mkdir` in a future release (https://github.com/nodejs/node/pull/21875), and I'd like to use that for `fse.mkdirp` in environments that support it, and Node's method will not support the `made` parameter.

RyanZim:
@manidlou FYI, I have labeled this with `feature-ensureDir`, since that's an alias for `mkdirs`, and deleted the `feature-mkdirs` label.

RyanZim:
Since we will be switching to Node's implementation in the future (https://github.com/jprichardson/node-fs-extra/issues/619), which does not support this parameter, we'll likely be removing it in the future. Upstream mkdirp has never patched this either. Closing this out.

PR #277, opened 2016-09-22
Mention defaults for clobber and dereference options

gaearon:
This wasn’t obvious to me.
Thought it’s worth documenting!


jprichardson:
Yes, thank you!

PR #280, opened 2016-09-28
Disable rimraf globbing

distracteddev:
Its important to pass the options through as ensuring symmetry between the `fs.extra#remove` and `rimraf` APIs means I don't need to include `rimraf` AND `fs-extra`, just for the sake of using `rimraf` with specific options.

One option that is particularly useful and quite commonly used is `disableGlob`.

### DIFF
--- a/lib/remove/index.js
+++ b/lib/remove/index.js
@@ -1,11 +1,12 @@
 var rimraf = require('rimraf')
 
 function removeSync (dir) {
-  return rimraf.sync(dir)
+  return rimraf.sync(dir, {disableGlob: true})
 }
 
 function remove (dir, callback) {
-  return callback ? rimraf(dir, callback) : rimraf(dir, function () {})
+  var options = {disableGlob: true}
+  return callback ? rimraf(dir, options, callback) : rimraf(dir, options, function () {})
 }
 
 module.exports = {


jprichardson:
I appreciate your thoughts here.

> Its important to pass the options through as ensuring symmetry between the fs.extra#remove and rimraf

This is assuming that I felt symmetry was important. I'm strongly considering removing `rimraf` since it feels like rimraf is getting bloated. 

I would accept a PR to set `{disableGlob: false}` to `rimraf` though since `fs-extra` doesn't allow this anyways.

distracteddev:
Heh, valid point re: symmetry.

distracteddev:
@jprichardson PR updated with globbing explicitly disabled

jprichardson:
Thank you. Please squash changes. Will merge after Oct 1 and subsequently publish `fs-extra@1.0.0`.

distracteddev:
@jprichardson done :)

jprichardson:
Thank you! I will merge after Oct 1st so that I can publish this along with dropping support for Node  v0.10.

jprichardson:
Actually, doesn't hurt for me to merge now. Will publish after Oct 1st :)

PR #282, opened 2016-10-06
Apply filter also on directories and symlinks for copySync()

nvdbleek:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fs-extra",
-  "version": "0.30.0",
+  "version": "0.30.1",
```
nvdbleek: sorry about that, I'll create a different branch and only cherry pick the first commit. Our automated build kept getting the original 0.30.0 version, so I updated it to work around it. But overlooked that it will automatically become part of my PR.

nvdbleek:
created new PR #283

Issue #240, opened 2016-04-22
walkSync request


rottmann:
Need a replacement for the deprecated wrench module `wrench.readdirSyncRecursive(path)`.

Currently "much" code and extra modules are necessary to retrieve only a dir list with walk in sync way.

sosspyker:
True we need a sync implmentation of walk() method. Gulp task with walk() (async) causes issues since some files may not be available for tasks down the line. Thx.

zaaack:
+1 👍

brentonhouse:
👍

johannasc:
+1

agnivade:
@jprichardson - I gave the incorrect PR (#286) tag to auto-close this issue. Would you mind manually closing this ?

PR #286, opened 2016-10-14
Added walkSync function

agnivade:


agnivade:
Closes #240

RyanZim:
@jprichardson Thoughts on this?

jprichardson:
I'm fine with keeping it in here. I really like how `klaw` simply just returns an async walker that is a ReadableStream. What are your thoughts @RyanZim?

RyanZim:
IDK, assigning to you.

jprichardson:
@agnivade what's your use case for `walkSync()`? i.e. why not just use the more efficient `walk()`?

agnivade:
I am working on a codebase which has used `sync` counterparts of all fs operations. Using the `async` versions will make it very hard to integrate. 

And moreover, if you are providing `sync` counterparts of all other `fs` operations, it only makes sense to add it for the `walk` functionality.

levithomason:
Same here, I'm working on a CLI that is loading a config file.  The entire bootstrap process and config loading is sync.  I need to read a dir sync and add all the paths found to some config.  If I go async here, the entire CLI bootstrapping needs to be rewritten to handle async.  It is also more complicated to handle async bootstrapping.

My use case was solved by [fs-readdir-recursive](https://github.com/fs-utils/fs-readdir-recursive), though I'd much rather not have another for one task.  I don't need the extra params provided by walk so my use case was a little more simple.

PR #289, opened 2016-10-16
(Closes #271) lib/util/utimes: properly close file descriptors in the event of an error

### LINKED ISSUE #271: Unclosed file handle on futimes error
The call to fs.close(fd) should be performed _before_ the call to return in this function:

https://github.com/jprichardson/node-fs-extra/blob/master/lib/util/utimes.js#L53

hhamilto:
I made one other change in utimes.js to ensure that the callback for `utimesMillis` is only called after the file has been closed.

### DIFF
--- a/lib/util/__tests__/utimes.test.js
+++ b/lib/util/__tests__/utimes.test.js
@@ -4,7 +4,9 @@ var os = require('os')
 var fs = require('fs')
 var fse = require(process.cwd())
 var semver = require('semver')
-var utimes = require('../utimes')
+var proxyquire = require('proxyquire')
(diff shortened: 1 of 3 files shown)


RyanZim:
@jprichardson Can you review this?

jprichardson:
Yep, it LGTM. How about you @RyanZim?

RyanZim:
Didn't look real close, but LGTM.

@jprichardson What's you're preferred merge method (squash, rebase, merge-commit)?

jprichardson:
Generally, I think `merge-commit` is good.

PR #291, opened 2016-10-26
Escape '$' in replacement string for async file copying

### LINKED ISSUE #242: Accidentally truncates file names ending with $$
So I thought it was a [browserify-hmr issue](https://github.com/AgentME/browserify-hmr/issues/29) but I looked into their tests and while I could produce an error I discovered via console log that `fsExtra.copy` was the point at which it incorrectly created a file. I'm now looking at fs-extra's tests to reproduce.

hhamilto:
Change made in /lib/copy/ncp.js to the string.replace in the onFile function

Fixes #242.

### DIFF
--- a/lib/copy/__tests__/copy.test.js
+++ b/lib/copy/__tests__/copy.test.js
@@ -89,6 +89,18 @@ describe('fs-extra', function () {
         })
       })
 
+      it('should copy to a destination file with two \'$\' characters in name (eg: TEST_fs-extra_$$_copy)', function (done) {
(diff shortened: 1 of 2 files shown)


RyanZim:
Fixes https://github.com/jprichardson/node-fs-extra/issues/242.

@jprichardson ?

hhamilto:
@RyanZim oops, sorry for that. I'll make sure to next time.

jprichardson:
Thank you @hhamilto!

Issue #292, opened 2016-10-26
Rewrite lib/copy/ncp.js


RyanZim:
This needs rewritten. It has several bugs and is overly complex. It was written for use with `fs` instead of `graceful-fs`, and contains a lot of concurrency-limiting code that is now unnecessary.

Once rewritten, it may be small enough to inline in lib/copy/copy.js; talk about that later.

I'll try to do this myself, but help welcome. Adding this to the v1.0.0 milestone, but may have to be pushed back to v2.0.0.

X-Refs: #233.

RyanZim:
@jprichardson Found why we need https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy/ncp.js#L150: http://stackoverflow.com/questions/34720794/create-directory-with-permission-in-node-js.

Advice here?

jprichardson:
I think that rewriting is a mistake for 1.0.0 and shouldn't even be considered even though it's bad legacy code. By all means, don't let me stop you, it's just in my long experience of writing code a rewrite is a bad idea 99% of the time.

RyanZim:
See #294

RyanZim:
@jprichardson Refs https://github.com/jprichardson/node-fs-extra/issues/83. This needs done sometime. (don't know when)

Would you consider a custom build/refactor based on https://github.com/davglass/cpr sometime in the future? Just a random idea, since it's one of the few well-maintained directory copiers.

Or do you want to roll your own, based on klaw?

manidlou:
So what is the decision on this? We need to decide what we want to do about this as `ncp.js` is the source of a lot of issues. I like @RyanZim idea about using native promise and avoid using counters.

@jprichardson what should we do about this?

jprichardson:
> Would you consider a custom build/refactor based on https://github.com/davglass/cpr sometime in the future? Just a random idea, since it's one of the few well-maintained directory copiers.

> Or do you want to roll your own, based on klaw?

I don't care much what it's based on at the moment, I just think whatever we need to do to get it to fulfill its implicit contract (re: work) is just fine by me! 😊

manidlou:
So @RyanZim how much are you progressed on rewriting or refactoring or whatever you wanna call it :smile: of `copy()`?! I rolled one using the good parts of what we've had already and replacing the bad parts such as counters with new stuff like native `Promise.all()` along with `Array.map()` to achieve better concurrency control.

[...]

PR #294, opened 2016-10-26
BREAKING: Refactor lib/copy/ncp.js

### LINKED ISSUE #233: please make sure the first argument in callback is error object
https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy/ncp.js#L237

### LINKED ISSUE #272: Copy clobber flag never err even when true
In the following code `err` is never not `null`:

```
fs.copy(sourcePath, targetPath, { clobber: true  }, (err) =>
{
    if (err)
        debugger
    fs.copy(sourcePath, targetPath, { clobber: true}, (err) =>
    {
        if (err)
            debugger
        fs.copy(sourcePath, targetPath, { clobber: false}, (err) =>
        {
            if (err)
                debugger
            fs.copy(sourcePath, targetPath, { clobber: false }, (err) =>
            {
                if (err)
                    debugger
            })
        })
    })
})
```

Maybe this is intended but then at least add the option for open flags so I can at least put `wx` and have it `err.code =  "EEXIST"`

Thanks,
Ryan

RyanZim:
@jprichardson Looking at https://github.com/jprichardson/node-fs-extra/issues/292, I decided to just do a fairly major refactor instead of a total rewrite.

Changes:

- Fix #233 - First argument passed to callback should be an Error object, not an array.
- Fix #272 - Should error when options.clobber is false and the destination exists.
- Remove options.limit, it is no longer needed with graceful-fs.

---

A test fails on windows, node v0.10:
```
1 failing
  1) ncp regular files and directories when using transform file descriptors are passed correctly:
     Uncaught AssertionError: null == true
      at C:\projects\node-fs-extra\lib\copy\__tests__\ncp\ncp.test.js:103:13
      at doneOne (C:\projects\node-fs-extra\lib\copy\ncp.js:226:16)
      at C:\projects\node-fs-extra\lib\copy\ncp.js:89:11
      at C:\projects\node-fs-extra\lib\copy\ncp.js:209:14
      at C:\projects\node-fs-extra\node_modules\graceful-fs\polyfills.js:267:18
      at Object.oncomplete (fs.js:108:15)
```
This is an undocumented feature, and we are dropping 0.10 support soon anyhow. Can you reproduce this error on your windows machine?

### DIFF
--- a/lib/copy/ncp.js
+++ b/lib/copy/ncp.js
@@ -16,18 +16,15 @@ function ncp (source, dest, options, callback) {
 
   var filter = options.filter
   var transform = options.transform
-  var clobber = options.clobber !== false
+  var clobber = options.clobber !== false // default true
   var dereference = options.dereference
   var preserveTimestamps = options.preserveTimestamps === true
-
-  var errs = null
 
   var started = 0
   var finished = 0
   var running = 0
-  // this is pretty useless now that we're using graceful-fs
-  // consider removing
-  var limit = options.limit || 512
+
+  var errored = false
 
   startCopy(currentPath)
 
@@ -49,11 +46,6 @@ function ncp (source, dest, options, callback) {
 
   function getStats (source) {
     var stat = dereference ? fs.stat : fs.lstat
-    if (running >= limit) {
-      return setImmediate(function () {
-        getStats(source)
-      })
-    }
     running++
     stat(source, function (err, stats) {
       if (err) return onError(err)
@@ -89,7 +81,11 @@ function ncp (source, dest, options, callback) {
             copyFile(file, target)
           })
         } else {
-          doneOne()
+          var err = new Error('EEXIST: ' + target + ' already exists.')
+          err.code = 'EEXIST'
+          err.errno = -17
+          err.path = target
+          onError(err)
         }
       }
     })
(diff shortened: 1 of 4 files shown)


jprichardson:
> @jprichardson Looking at #292, I decided to just do a fairly major refactor instead of a total rewrite.

Ah, thank goodness :)

> A test fails on windows, node v0.10:

The great news is that we don't need Node v0.10 anymore. It's unsupported as of Oct 1st.

lib/copy/ncp.js
```diff
@@ -133,7 +130,7 @@ function ncp (source, dest, options, callback) {
   }
 
   function onDir (dir) {
-    var target = dir.name.replace(currentPath, targetPath)
+    var target = dir.name.replace(currentPath, targetPath.replace('$', '$$$$')) // escapes '$' with '$$'
```
hhamilto: Thanks for adding that :) I should have caught it when I changed onFile.

RyanZim: No problem!

lib/copy/ncp.js
```diff
@@ -89,6 +81,11 @@ function ncp (source, dest, options, callback) {
             copyFile(file, target)
           })
         } else {
+          var err = new Error('EEXIST: ' + target + ' already exists.')
+          err.code = 'EEXIST'
+          err.errno = -17
+          err.path = target
+          onError(err)
```
RyanZim: @hhamilto Good catch! I meant to remove the `doneOne` call below, but somehow missed it.

RyanZim: Fixed! No good way to unit-test this, this really should be rewritten using promises; someday...

RyanZim:
@jprichardson Appveyor is passing now. How does that work?

PR #295, opened 2016-10-26
Depreciate using regular expressions for copy's filter option

### LINKED ISSUE #239: Depreciate regular expressions for copy's filter
Regular Expression filters in [ncp.js:38](https://github.com/jprichardson/node-fs-extra/blob/141422ee384fda83e3e1d4af2727bc9dca799f68/lib/copy/ncp.js#L38) are run against the full source path, i.e.:

`/Users/martin/Private/projects/nodeschool/git-it-electron/node_modules/eslint/node_modules/uglify-js/node_modules/source-map/lib/util.js`

If I want in the folder `git-it-electron` to exclude the `eslint` folder I need to know that I am in the `/Users/martin/Private/projects/nodeschool/git-it-electron` folder when I do that. If I test for `/eslint/` then It would exclude the folder properly but If by some accident I would put the project in a folder called:

`/Users/eslint/Private/projects/nodeschool/git-it-electron`

It would ignore all of the files in my project 😢 . This means that someone would need to implement the file matching with that in mind. Given that `ncp.js` already [considers the working directory](https://github.com/jprichardson/node-fs-extra/blob/141422ee384fda83e3e1d4af2727bc9dca799f68/lib/copy/ncp.js#L13) I think it would be specially awesome if

   A. there would be the opportunity to have filters that work relatively to the working directory
   B. there would be the opportunity to specify where the working directory is (overriding the good `process.cwd()` default)

Additionally it would be super awesome if the filter would allow [glob](https://www.npmjs.com/package/glob) file patterns instead of regexp patterns because they are made for files.

_Note: I am thinking about the API a little and think it would be good to have an `include` and `exclude` option instead of filters. (Considering #65)_

RyanZim:
Fixes #239 

Attn: @jprichardson

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Options:
 - clobber (boolean): overwrite existing file or directory, default is `true`.
 - dereference (boolean): dereference symlinks, default is `false`.
 - preserveTimestamps (boolean): will set last modification and access times to the ones of the original source files, default is `false`.
-- filter: Function or RegExp to filter copied files. If function, return true to include, false to exclude. If RegExp, same as function, where `filter` is `filter.test`.
+- filter: Function to filter copied files. Return `true` to include, `false` to exclude. This can also be a RegExp, however this is depreciated (See [issue #239](https://github.com/jprichardson/node-fs-extra/issues/239) for background).
 
 Sync: `copySync()`
 


jprichardson:
Thanks!

Would you be willing to add one more change to this checking if `filter instanceof RegExp` and throw a `console.warn`? 😁

RyanZim:
I can, I was thinking perhaps that should be added in a later release to avoid raising too many tempers regarding `console.warn` in a library. IDK

RyanZim:
@jprichardson So do we `console.warn` or not?

jprichardson:
> @jprichardson So do we console.warn or not?

Yes, I think it's a good idea to be naggy so that devs change and are aware.

PR #297, opened 2016-10-27
Warn about using preserveTimestamps on 32-bit node

### LINKED ISSUE #269: Copying with preserveTimeStamps: true is inaccurate using 32bit node
I have been running the fs-extra tests via npm test on various platforms, and have been getting failures on the 32bit versions of node.  I wasn't sure if millisecond precision would be supported or not at 32bits, and it should be pointed out that the underlying OS is 64bit, but we do get back valid millisecond precision timestamps using 32bit node, it's just not accurate once we copy a file and ask to preserve timestamps.

This can be reproduced using this small snippet of code -

```
var fs = require('fs')
var fse = require('fs-extra')
var from = './forcopy' 
var to = './copied'

fse.copySync(from, to, {preserveTimestamps: true})
var fromStat = fs.statSync(from)
var toStat = fs.statSync(to)
console.log('From stat: ' , fromStat.mtime.getTime())
console.log('To stat: ' , toStat.mtime.getTime())
```

if I then `touch` the `forcopy` file and then run with 64bit node

```
$ which node
~/node-v4.5.0-linux-x64/bin/node
$ node copyTest
From stat:  1472712683269
To stat:  1472712683269
```

and then with 32bit node

```
$ which node
~/node-v4.5.0-linux-x86/bin/node
$ node copyTest
From stat:  1472712683269
To stat:  1472712683208
```

I can replicate this problem on rhel7, sles12 and ubuntu 14-04

RyanZim:
Fixes #269

### DIFF
--- a/lib/copy-sync/copy-sync.js
+++ b/lib/copy-sync/copy-sync.js
@@ -18,6 +18,12 @@ function copySync (src, dest, options) {
 
   options.filter = options.filter || function () { return true }
 
+  // Warn about using preserveTimestamps on 32-bit node:
+  if (options.preserveTimestamps && process.arch === 'ia32') {
+    console.warn('fs-extra: Using the preserveTimestamps option in 32-bit node is not recommended;\n' +
+    'see https://github.com/jprichardson/node-fs-extra/issues/269')
+  }
+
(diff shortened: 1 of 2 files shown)


jprichardson:
Thank you!

PR #300, opened 2016-10-27
Inline Rimraf

RyanZim:
We aren't using rimraf's glob support, and glob is a rather large dependency.

Changes:

- Remove glob support from rimraf
- Remove move()'s dependency on rimraf; use fs-extra.remove instead
- Make rimraf a devDependency; it is used in the tests

---

**Original comment:**

@jprichardson This is my WIP of inlining rimraf. This needs rebased, **do not merge**.

rimraf has functionality built in to handle alternate `fs` modules. Currently, there is no way for the end-user to leverage that functionality. I could simplify the code by removing that, however, I don't want to have to add it back. Will we be adding custom `fs` module support or glob support to `fs-extra` first?

### DIFF
--- a/lib/remove/rimraf.js
+++ b/lib/remove/rimraf.js
@@ -1,0 +1,301 @@
+module.exports = rimraf
+rimraf.sync = rimrafSync
+
+var assert = require('assert')
+var path = require('path')
+var fs = require('graceful-fs')
+
+var isWindows = (process.platform === 'win32')
+
+function defaults (options) {
+  var methods = [
+    'unlink',
+    'chmod',
+    'stat',
+    'lstat',
+    'rmdir',
+    'readdir'
+  ]
+  methods.forEach(function (m) {
+    options[m] = options[m] || fs[m]
+    m = m + 'Sync'
+    options[m] = options[m] || fs[m]
+  })
+
+  options.maxBusyTries = options.maxBusyTries || 3
+}
+
+function rimraf (p, options, cb) {
+  if (typeof options === 'function') {
+    cb = options
+    options = {}
+  }
+
+  assert(p, 'rimraf: missing path')
+  assert.equal(typeof p, 'string', 'rimraf: path should be a string')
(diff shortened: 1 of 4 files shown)


jprichardson:
> Will we be adding custom fs module support 

This will come first.

RyanZim:
OK, thanks.

lib/remove/rimraf.js
```diff
@@ -0,0 +1,314 @@
...
+        busyTries++
+        var time = busyTries * 100
+        // try again, with the same exact callback as this one.
+        return setTimeout(function () {
+          rimraf_(p, options, CB)
+        }, time)
+      }
+
+      // this one won't happen if graceful-fs is used.
+      if (er.code === 'EMFILE' && timeout < options.emfileWait) {
```
RyanZim: @jprichardson This is a custom handler for `EMFILE` errors, only needed if you are using a custom `fs`. This brings up a question; when we add custom `fs` module support, will we add checks for `EMFILE` errors (where possible) across `fs-extra`?

jprichardson: > will we add checks for EMFILE errors (where possible) across fs-extra?

I don't think so, at least not in the short-term.

RyanZim: OK, then I guess I'll remove these checks here.

jprichardson: > OK, then I guess I'll remove these checks here.

I don't it hurts to keep them. Whatever is the shortest path and saves you time!

RyanZim: @jprichardson I thought I would remove them in interests of a consistent api. If they are left here, `remove` will handle `EMFILE` errors while the rest of the api won't (when using custom `fs` modules).

RyanZim:
@jprichardson Done & rebased. Anything else?

---

What else needs done for v1.0.0?

jprichardson:
Thanks!

> @jprichardson Done & rebased. Anything else?

I think removing `path-is-absolute` as a dependency since we no longer support Node v0.10 is the last item.

Issue #199, opened 2015-12-08
fs.copySync fails if user does not own file


bdcribbs:
fs.copySync fails when the user running the script does not own the destination file, even when the user actually has sufficient permissions to perform the action.

The error occurs if the file exists, and clobber is true.  In that case, copySync attempts to chmod the file prior to unlinking it (prior to performing the copy).  Since the user doesn't own the file, the chmod fails with EPERM, even though the operation could have succeeded -- provided the user has sufficient permissions to unlink the file and recreate it (which depends on the permissions of the directory the file is in -- not the file itself).

The line in question is here:
https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy-sync/copy-file-sync.js#L12

jprichardson:
Thanks so much! Would you be willing to add a test?

bdcribbs:
I'm not sure how we'd write such a test, since it requires creating a file as one user account and then attempting to manipulate it with another user account.

jprichardson:
Good point. Maybe throw that line in a try/catch block?

bdcribbs:
A try/catch would work.  Another point of view is that attempting to copy over the contents of a read-only file really *should* fail, because that's what read-only means.

jprichardson:
> Another point of view is that attempting to copy over the contents of a read-only file really should fail, because that's what read-only means.

Hah. Excellent point as well.

So for a test, couldn't we just set a couple of files to be read only and try to copy over them and assert for an error? The error should have a descriptive message indicating that the destination is read-only.

bdcribbs:
That works, and also matches how I would expect it to function (I was somewhat surprised when I found out copySync was calling chmod in the first place).

It's worth considering that this would revert the change that was originally made for #183, but in my opinion, it's more appropriate for the caller to chmod the file themselves if they need to.

If you'd like I could throw together a pull request, but I didn't want to revert #183 without your agreement.

jprichardson:
@bdcribbs thanks for referencing the issue. In context it makes more sense. Don't you think that when clobber (aka overwrite) is set to `true`, it should overwrite regardless of whether the file is read only?

Pinging: @bartland for his feedback.

bdcribbs:
I see. I thought clobber was for overwriting files that already exist (EEXIST), not EPERM.  Now the behavior makes more sense.

[...]

PR #301, opened 2016-10-27
Remove chmod call from copySync

RyanZim:
This should resolve https://github.com/jprichardson/node-fs-extra/issues/199.


jprichardson:
Thanks!

Issue #180, opened 2015-09-25
Filter option doesn't work the same way in copy and copySync


stefk:
`copySync` applies the filter only to files, while `copy` applies the filter to every entry.

So if you have the following directory structure:

```
src
  foo
    baz
      file1 
  bar
    baz
      file2
```

calling `copy` with `{ filter: /baz/ }` will produce:

```
dest
  foo
  bar
```

wheras `copySync` will do:

```
dest
  foo
    baz
  bar
    baz
```

The former is obviously what's expected.

jprichardson:
Thanks for pointing this out :)

jehendricks:
I've been experiencing the same problem and I think I may have found what is going on.  In copy-sync, the check for filter that sets performCopy is only checked inside the stats.isFile() conditional.  I moved the setting of performCopy outside the conditional then wrapped the whole conditional (isFile, isDirectory, isSymbolicLink) in a performCopy conditional.  That worked, but of course I may have broken some other internal code functionality in the process.  Here's my code fragment if it's any help:

ar performCopy = false

  if (options.filter instanceof RegExp) performCopy = options.filter.test(src)
  else if (typeof options.filter === 'function') performCopy = options.filter(src)

  if (performCopy) {
    if (stats.isFile()) {
      if (!destFolderExists) mkdir.mkdirsSync(destFolder)
      copyFileSync(src, dest, {clobber: options.clobber, preserveTimestamps: options.preserveTimestamps})
    } else if (stats.isDirectory()) {
      if (!fs.existsSync(dest)) mkdir.mkdirsSync(dest)
      var contents = fs.readdirSync(src)
      contents.forEach(function (content) {
        var opts = options
        opts.recursive = true
        copySync(path.join(src, content), path.join(dest, content), opts)
      })
    } else if (options.recursive && stats.isSymbolicLink()) {
      var srcPath = fs.readlinkSync(src)
      fs.symlinkSync(srcPath, dest)
    }
  }

RyanZim:
@jprichardson Should this added to the v1.0.0 milestone?

RyanZim:
@jprichardson ping?

jprichardson:
> @jprichardson Should this added to the v1.0.0 milestone?

Ideally yes, but if it's going to be a lot of work, no.

RyanZim:
OK, should document this though.

RyanZim:
Fixed by #324.

PR #303, opened 2016-10-28
Document copySync filter inconsistency

RyanZim:
Refs https://github.com/jprichardson/node-fs-extra/issues/180#issuecomment-257016134


jprichardson:
Thanks!

PR #305, opened 2016-10-29
Re-add Coveralls

RyanZim:
@jprichardson Not sure if this is the best way, as it runs the tests twice. Other ways:

- Only run `coveralls` on the `LINT` build.
- Run the coverage script as the main script, and modify the coveralls script to only use the existing coverage report.
- Write a custom travis script in JS that runs tests for every platform and runs `standard` and `coveralls` on only one version.

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ matrix:
 env:
   - TEST_SUITE=unit
 script: npm run-script $TEST_SUITE
+after_success:
+  - npm run coveralls


jprichardson:
> Thoughts?

I think the simplest path is the best at least in the short-term, so this is probably best:

> Only run coveralls on the LINT build.

RyanZim:
OK, I'll try to do that when I get time.

RyanZim:
Oops, coverage reports are inaccurate: https://coveralls.io/github/jprichardson/node-fs-extra. Istanbul is figuring coverage for the `__test__` files. Need to set istanbul to ignore these.

PR #306, opened 2016-10-29
Update devDeps, fix lint error

RyanZim:
@jprichardson Ever think of using @greenkeeperio-bot so I don't have to do this? I use it on all of my projects and really like it. Not only does it update dependencies outside your version range, it also checks to make sure none of your dependencies break semver.

My only gripe is the commit messages: `chore(package): update eslint to version 3.8.0`. I always use the squash-merge function so I can reword the commit message to `Update eslint to version 3.8.0`.

https://greenkeeper.io/

### DIFF
--- a/lib/util/__tests__/utimes.test.js
+++ b/lib/util/__tests__/utimes.test.js
@@ -37,7 +37,7 @@ describe('utimes', function () {
       // does anyone use FAT anymore?
       /* if (process.platform === 'win32') {
         assert.equal(res, true)
-      }*/
+      } */
       // fails on appveyor... could appveyor be using FAT?
 
       // this would fail if ext2/ext3
--- a/package.json
+++ b/package.json
@@ -39,15 +39,15 @@
   },
   "devDependencies": {
     "coveralls": "^2.11.2",
-    "istanbul": "^0.3.5",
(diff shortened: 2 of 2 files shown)


jprichardson:
Thanks for this!

> Ever think of using @greenkeeperio-bot so I don't have to do this? 

I'm not a fan of it as I think that this promotes updating dependencies for the sake of updating dependencies as a good thing. This is not to say that I couldn't warm up to it, but I'm not ready for it yet.

RyanZim:
> I think that this promotes updating dependencies for the sake of updating dependencies as a good thing

Not sure what you mean by that, it only updates your deps when the latest version is not covered by your current version range. Mind elaborating?

jprichardson:
> Not sure what you mean by that, it only updates your deps when the latest version is not covered by your current version range. Mind elaborating?

Sure. There's a unforeseen cost when updating dependencies. You rarely don't know what it is up front. Greenkeeper and the like promote that to have stable and usable software, you must update your dependencies. Software developers feel good about this - one more thing to check off the list. I'm saying that these services don't show the unforeseen cost. I believe it's the developer duty and responsibility to be judicious about updating and the need to update.

PR #307, opened 2016-10-31
Fix coverage

RyanZim:
- istanbul shouldn't include test files in the coverage report
- Run coveralls only on the lint build in Travis CI

Coverage is pretty low, mainly due to window-specific code. How would coveralls handle it if we ran it on appveyor as well? Coverage would be higher there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,4 +15,4 @@ env:
   - TEST_SUITE=unit
 script: npm run-script $TEST_SUITE
 after_success:
-  - npm run coveralls
+  - if [ $TEST_SUITE = lint ]; then npm run coveralls; fi
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   },
   "main": "./lib/index",
   "scripts": {
-    "coverage": "istanbul cover test.js",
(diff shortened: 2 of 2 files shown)


jprichardson:
> How would coveralls handle it if we ran it on appveyor as well? 

Very good question. I'm not sure at the moment. But I really appreciate you fixing this!

RyanZim:
BTW:

> Source Not Available

> The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API.

@jprichardson?

jprichardson:
Is this AppVeyor? I just logged on and checked... everything looks okay. If it's coveralls causing fits, I'm fine with switching to just plain NYC or something.

RyanZim:
@jprichardson This message is at https://coveralls.io/builds/8588465/source?filename=lib%2Fmkdirs%2Fmkdirs-sync.js. Coveralls works fine, except for the fact that you can't see the line-by-line coverage reports online. Sorry, should have made that clearer.

jprichardson:
No problem, done!

PR #313, opened 2016-11-14
Test that remove() ignores glob characters.

binki:
After looking at rimraf, I began to fear that node-fs-extra, which
is based on rimraf, might also automatically glob things. I find
this confusing when the API is being used as an overlay of the
node built-in fs module which never expands globs. This test
explicitly verifies that the remove() function does not follow
glob characters.

### DIFF
--- a/lib/remove/__tests__/remove.test.js
+++ b/lib/remove/__tests__/remove.test.js
@@ -83,7 +83,14 @@ describe('remove', function () {
 
     it('shouldn’t delete glob matches', function (done) {
       var file = path.join(TEST_DIR, 'file?')
-      fs.writeFileSync(file, 'hello')
+      try {
+        fs.writeFileSync(file, 'hello')
+      } catch (ex) {
+        if (ex.code === 'ENOENT') {
+          return this.skip('Windows does not support filenames with ‘?’ or ‘*’ in them.')
+        }
+        throw ex
+      }
 
       var wrongFile = path.join(TEST_DIR, 'file1')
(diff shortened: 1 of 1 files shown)


RyanZim:
@jprichardson What's the issue with the windows tests?

We are using a custom version of rimraf with glob support removed; but I guess extra tests don't hurt. @jprichardson What are your thoughts on this?

binki:
@RyanZim The Windows tests always fail because of https://github.com/isaacs/node-graceful-fs/issues/98 .

I was also thinking that *if* you ever decided to unbundle rimraf these tests could help verify that the globs don’t match, and I do like to verify that they’re not factored regardless of how the implementation changes.

jprichardson:
> We are using a custom version of rimraf with glob support removed; but I guess extra tests don't hurt. @jprichardson What are your thoughts on this?

Yep, I like extra tests as well.

> @RyanZim The Windows tests always fail because of isaacs/node-graceful-fs#98 .

Can you fix this PR to make the tests pass?

binki:
> Can you fix this PR to make the tests pass?

I’m going to create a new PR to “fix” that particular test by increasing the timeout for that one test to 1m30s if I can figure that out.

RyanZim:
@binki Do `this.timout = 90 * 1000` at the top of the `test()` function.

RyanZim:
@binki Can you rebase this onto `master`?

binki:
History rewritten.

jprichardson:
Thanks @binki!

PR #315, opened 2016-11-14
Fix move clobber tests to work around graceful-fs bug.

binki:
When trying to rename a directory over another directory, graceful-fs-4.1.10
(maybe other versions?) takes 1 minute to fail the call with an error.
This causes the test to timeout. The workaround is to increase the timeout
of that one test to 1m30s.

See https://github.com/isaacs/node-graceful-fs/issues/98


binki:
Yay, first appveyor succeeded: https://ci.appveyor.com/project/jprichardson/node-fs-extra/build/1.0.404/job/ug8a160wvg2q0o3m

jprichardson:
Thank you @binki!

PR #317, opened 2016-11-15
Adding copy sync test for src file without write perms

weexpectedTHIS:
Adding test coverage to cover case I thought previously was a bug


weexpectedTHIS:
It doesn't cover more lines of code, covers more cases. You can reject if you don't want. Just didn't want my effort to replicate the previous problem I had to go to waste.

jprichardson:
@weexpectedTHIS thanks!

PR #324, opened 2016-12-05
copySync() should apply filter to directories like copy()

### LINKED ISSUE #180: Filter option doesn't work the same way in copy and copySync

RyanZim:
Fixes #180 

Had to change a incorrectly written test.

@jprichardson ?


jprichardson:
Thanks!

Issue #326, opened 2016-12-20
fs.copy fails with chmod error when disk under heavy use


nicks:
Sometimes, when the disk is under heavy use, we see fs.copy fail with the following error:
```
[Error] ENOENT: no such file or directory, chmod '/path/to/dir.js'
```

We tracked down the error to this line: https://github.com/jprichardson/node-fs-extra/blob/1.x/lib/copy/ncp.js#L111
```
    writeStream.once('finish', function () {
      fs.chmod(target, file.mode, function (err) {
        if (err) return onError(err)
```

My current hypothesis is that the stream `finish` event is just for generic stream flush, and doesn't guarantee that the underlying filesystem operations have completed. If I change it to 'close', the issue goes away.

That said, I've very vague on the intended semantics of nodejs stream events, so I might be on the totally wrong path here.

jprichardson:
I believe you are correct. Feel like submitting a PR? We'll be releasing 2.0 after Jan 1st and this would make in time for that release.

nicks:
OK

RyanZim:
Fixed in https://github.com/jprichardson/node-fs-extra/pull/327 (Thanks @nicks!). Should be released early next year.

popod:
It seem that this error occurs always on 2.0.0 :(

![error](https://cloud.githubusercontent.com/assets/3807312/22714064/31711618-ed8b-11e6-87d5-eebee4cf7746.jpg)

RyanZim:
@popod If this issue still exists, please open a new issue with a minimal, complete, and verifiable example. Thanks!

PR #327, opened 2016-12-20
Use writeStream 'finish' event instead of 'close'

nicks:
Fixes https://github.com/jprichardson/node-fs-extra/issues/326


nicks:
oops sorry! ya it should be better now.

PR #330, opened 2016-12-24
BREAKING: Do not error when copy destination exists & clobber: false

### LINKED ISSUE #321: fs.copy() with clobber set to false return EEXIST error
In version 1.0.0, `fs.copy()` with clobber set to false return an EEXIST error.

I think this is a bug because we explicitly set clobber to false (overwrite existing file), so this should not return an error !

RyanZim:
Add errorOnExist option for users that want an error

Resolves #321 

@jprichardson What do you think of the error message?

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,8 @@ Methods
 Copy a file or directory. The directory can have contents. Like `cp -r`.
 
 Options:
-- clobber (boolean): overwrite existing file or directory, default is `true`.
+- clobber (boolean): overwrite existing file or directory, default is `true`. _Note that the copy operation will silently fail if you set this to `false` and the destination exists._ Use the `errorOnExist` option to change this behavior.
+- errorOnExist (boolean): when `clobber` is `false` and the destination exists, throw an error. Default is `false`.
 - dereference (boolean): dereference symlinks, default is `false`.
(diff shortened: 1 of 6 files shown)


RyanZim:
Hold off on this, copySync needs edited too.

RyanZim:
copySync updated.

RyanZim:
@jprichardson ?

RyanZim:
@jprichardson I think this is merge-ready, just want you to approve the error message first.

jprichardson:
Awesome, thanks!

callodacity:
Just a thought - do you think returning the same style error messages as `fs` returns would be better suited since this module is meant as a drop-in replacement? 

Meaning that instead of `${path} already exists` we could return an `EEXIST` error?

RyanZim:
@callodacity We've discussed this before, I don't think replicating fs errors is a good practice.

1. It's impossible to fully replicate a true native error.
1. If the error is coming from fs-extra, that should be clear, since fs-extra might have a bug.

callodacity:
@RyanZim both good points, and I agree with you completely. 

However, the only thing I see as an issue is that in order to check if the error thrown is the `already exists` error, one would have to search the string. I just thought it may be simpler is an error code or number were added (so users could just check for that rather than searching the string for `already exists`.

PR #343, opened 2017-01-13
Add klaw-sync link to readme

### LINKED ISSUE #338: Remove walk() and walkSync()
@RyanZim and I have decided to remove `walk()` and `walkSync()`. 

This was not an easy decision, but the long-term vision of this library is to provide Node.js programmers with an easy to use, batteries-included way to do file system operations. While we've succeeded up to this point, @RyanZim and I must be judicious about our resources. 

I think there's a strong case to be made for `walk()` to be in the library, but the Streams interface may not belong - perhaps generators are a better way to accomplish this? 

ES2015 and ES2017 bring generators and async/await. This ushers in a new era of writing JavaScript code. I believe that we'll see a decline in using `sync()` methods over the years. However, before too many people establish dependencies upon the synchronous code, it should be removed sooner rather than later. We will start with `walkSync()`. The other sync methods will stay with us for awhile (at least one year).

`walkSync()` joined `fs-extra` for the `1.0.0` release (two months ago) and with `2.0.0` coming soon, it's going to be removed along with `walk()`. Users of `walk()`have a drop-in replacement with [klaw](https://github.com/jprichardson/node-klaw). But we don't want to leave users of `walkSync()` in the dust (I don't think there are many at this point), but @mawni / @agnivade would you be interested in creating a `klaw-sync` or `walk-sync` library for us to point to users?

mawni:
This resolves #338.

It adds https://github.com/mawni/node-klaw-sync link to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,9 @@ Methods
 
 **NOTE:** You can still use the native Node.js methods. They are copied over to `fs-extra`.
 
-### What happened to `walk()`?
+### What happened to `walk()` and `walkSync()`?
 
-It was removed from `fs-extra` in v2.0.0. If you need the functionality, `walk` is available as a separate package, [`klaw`](https://github.com/jprichardson/node-klaw).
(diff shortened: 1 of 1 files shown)


jprichardson:
Thanks @mawni!

Issue #355, opened 2017-02-13
es6 refactoring


JPeer264:
Since just Node v4+ is supported wouldn't it be could to rewrite everything into ES6?

RyanZim:
Yeah, Node 4+ supports most, but not all ES6 features. Given that this is a large codebase, porting to ES6 will take some time.

Currently, I've had a fairly passive approach to this; when I worked on a certain part of the codebase, I would port it to ES6.

If you want to work on this, thought, PRs welcome.

---

@jprichardson Do you have ES6 stylistic preferences? (i.e. should `var` be used)

jprichardson:
> @jprichardson Do you have ES6 stylistic preferences? (i.e. should var be used)

We should avoid var and use let/const if we can.

manidlou:
@jprichardson @RyanZim regarding the codebase, both internally and for the exported functions, do we want to keep the callback style or we want to change that to promise-based?

RyanZim:
@manidlou: @jprichardson & I have discussed promise support via email in the past. It's something we definitely want to do sometime, but we want to do it right. That means:

- We need to support both promises & callbacks; if a callback is passed, use it; otherwise return a promise.
- We need to wrap the native `fs` methods as well.

We also want a solution that will allow incremental porting of the internals to promises (i.e. our solution must be able to wrap both promise and callback functions).

I've been planning on attacking this as soon as I get time; I wanted to do it for v2, but it didn't happen.

PR #356, opened 2017-02-14
ES6 refactoring

JPeer264:
I refactored some files to ES6 according to #355 . Much things are not available in Node v4.

`'use strict'` is a must in Node v4 if `let` is used. For consistency I also put it although no `let` is used.

### DIFF
--- a/lib/copy-sync/copy-file-sync.js
+++ b/lib/copy-sync/copy-file-sync.js
@@ -1,26 +1,28 @@
-var fs = require('graceful-fs')
+'use strict'
 
-var BUF_LENGTH = 64 * 1024
-var _buff = new Buffer(BUF_LENGTH)
+const fs = require('graceful-fs')
+
+const BUF_LENGTH = 64 * 1024
+const _buff = new Buffer(BUF_LENGTH)
 
 function copyFileSync (srcFile, destFile, options) {
-  var overwrite = options.overwrite
-  var errorOnExist = options.errorOnExist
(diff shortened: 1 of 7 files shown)


RyanZim:
`'use strict'` isn't required for `let` in Node 4 (http://node.green/#let), but it's a good practice anyway.

RyanZim:
@JPeer264 Thanks for doing this!

If you feel like refactoring other parts (including tests), you are more than welcome to. :smile_cat:

JPeer264:
> 'use strict' isn't required for let in Node 4 (http://node.green/#let), but it's a good practice anyway.

@RyanZim I read this too but my shell output told me something else. Also it is just with `let`:

```sh
let mkdir = require('../mkdirs')
^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    ...
```

Thanks. I will first finish all other parts (including README) and start with all tests, if that is ok 👍

RyanZim:
> my shell output told me something else

Strange. My test:

```
$ node -v
v4.5.0
$ node
> let a = 4
undefined
> console.log(a)
4
undefined
```

JPeer264:
In the shell directly it works fine for me as well. But not if the you run it from a file:

```sh
$ node -v
v4.7.3
$ echo "let a = 1" >> foo.js && node foo.js
```

RyanZim:
Huh, never knew that.

PR #359, opened 2017-02-17
Document the return value of ensureDir

### LINKED ISSUE #349: Feature request: Provide created path for fs.ensure*()
It would in some cases be useful to get feedback at which point an `fs.ensure*()` method did start to create a folder structure.

Two examples will probably clarify this best:

```js
// Existing path: /my/example/path

// Example 1
fs.ensureDir('/my/example/path/with/some/additions', function (err, created) {
  assert(created === '/my/example/path/with')
})

// Example 2
fs.ensureDir('/my/example', function (err, created) {
  assert(created === null)
})
```

My concrete use case is that I have to run `fs.ensureDir()` with root privileges and then set the owner back to the user who ran the command.

loilo:
This feature is implemented in mkdirp and verified in the mkdirp tests. Just adding it to documentation.
Fixes #349

### DIFF
--- a/docs/ensureDir.md
+++ b/docs/ensureDir.md
@@ -18,3 +18,18 @@ fs.ensureDir(dir, function (err) {
   // dir has now been created, including the directory it is to be placed in
 })
 ```
+
+---
+
+The callback receives a second argument, containing the path to the first directory that was created in the process. If no directories were made, this will be `null`.
+```js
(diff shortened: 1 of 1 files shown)


RyanZim:
Will probably be dropping this due to https://github.com/jprichardson/node-fs-extra/issues/619; so closing this out.

PR #360, opened 2017-02-17
ES6 refactor

JPeer264:
(#355 )  I did some more refactoring. Now everything except `move/index`, `copy/ncp` and `remove/rimraf` are as much as possible refactored to ES6. I wasn't sure on the skipped files, as they are from another library or got a comment with something like `needs rewrite` or similar.

### DIFF
--- a/lib/ensure/symlink-paths.js
+++ b/lib/ensure/symlink-paths.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const path = require('path')
-// path.isAbsolute shim for Node.js 0.10 support
 const fs = require('graceful-fs')
 
 /**
--- a/lib/ensure/symlink-paths.js
+++ b/lib/ensure/symlink-paths.js
@@ -1,6 +1,8 @@
-var path = require('path')
+'use strict'
+
+const path = require('path')
 // path.isAbsolute shim for Node.js 0.10 support
-var fs = require('graceful-fs')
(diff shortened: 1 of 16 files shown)


lib/ensure/symlink-paths.js
```diff
@@ -1,6 +1,8 @@
-var path = require('path')
+'use strict'
+
+const path = require('path')
 // path.isAbsolute shim for Node.js 0.10 support
```
RyanZim: Sort of unrelated, but this comment should have been removed earlier, wouldn't hurt to remove it here.

RyanZim:
I'm fine with doing the skipped files, all of them are already significantly modified from their original state.

If we decide to make changes in https://github.com/jprichardson/node-fs-extra/issues/292, you could skip `ncp`; not sure what we're doing there yet.

JPeer264:
@RyanZim I am on `move/index` right now. As there are many functions inside I would suggest to outsource them into own files and require them separately into `index.js`. How you think about it?

RyanZim:
`move/index` is only 158 lines, I wouldn't split it personally, perhaps @jprichardson differs.

BTW, don't bother with `copy/ncp`; I'm working on that right now.

manidlou:
@RyanZim is this ready to merge?

RyanZim:
Merged. @JPeer264 Thanks again for all your work here.

PR #362, opened 2017-02-20
Refactor move / rimraf to ES6

JPeer264:
(#355)
`move`:

I renamed the function from `mv` to `move`, it shouldn't be a breaking change

`rimraf`:

Here I put `module.exports` to the end of the file, as it is the same in all other files. Also I refactored one line `if`-statements, also with the same reason that it is the same in all other files.

---
I skipped `copy/ncp` as discussed in https://github.com/jprichardson/node-fs-extra/pull/360#issuecomment-280857354. Now I will start with the tests, unless I missed anything. I will make a pull request for each folder (`copy`, `copy-sync`, ...) if that is ok for you.

### DIFF
--- a/lib/move/index.js
+++ b/lib/move/index.js
@@ -1,23 +1,25 @@
+'use strict'
+
 // most of this code was written by Andrew Kelley
 // licensed under the BSD license: see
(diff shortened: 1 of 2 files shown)


jprichardson:
Thanks @JPeer264!

PR #363, opened 2017-02-21
lib/copy/ncp.js: use finish event instead of close in writestream when copying file

### LINKED ISSUE #357: fs.move fails with chmod error when disk under heavy use
Using fs-extra v2.0.0 I noticed that `fs.move`, which uses `/lib/copy/ncp` under the hood, fails with a `chmod` error when the disk is under heavy use.
```
{ Error: ENOENT: no such file or directory, chmod '/the/path-a/file'
  errno: -2,
  code: 'ENOENT',
  syscall: 'chmod',
  path: '/the/path-b/file' }
```

In this case I'm moving a very large directory of a few files.
I'm using [pify](https://github.com/sindresorhus/pify) on `fs.move`:
```js
const binary = (func) => (a, b) => func(a, b)
const move = binary(pify(fs.move))
// ...
move('/the/path-a', '/the/path-b')
```

With this error on the move operation what ends up happening is that `path-a` will still exist with one or so of its files left but `path-b` is created and contains all files *(at least it contains a copy of the file remaining in `path-a`)*.

Related to #326. \cc @nicks.

manidlou:
This should resolve #357. I provided the argument there. Based on the node docs, I believe  for writestreams the `finish` event is the correct one to use not the `close` event.

### DIFF
--- a/lib/copy/ncp.js
+++ b/lib/copy/ncp.js
@@ -109,7 +109,7 @@ function ncp (source, dest, options, callback) {
       })
     }
 
-    writeStream.once('close', function () {
+    writeStream.once('finish', function () {
       fs.chmod(target, file.mode, function (err) {
         if (err) return onError(err)
         if (preserveTimestamps) {


manidlou:
I am closing this as a result of discussion in #357. Accordingly it is more appropriate to keep the `close`event and not changing it back to `finish`.

PR #365, opened 2017-02-22
Add dest as an additional parameter to filter option in copy and copySync

### LINKED ISSUE #351: Additional param for options.filter() on copy()
At the moment when using `copy()`, the `options.filter()` function receives the source `path` as it's only param. It would be nice to have both the source and destination paths:

```javascript
fs.copy(srcFolder, destFolder, {
  filter: (srcPath, destPath) => {
    return true
  }
}, cb)
```

This would make checking last mod dates much easier.

manidlou:
This resolves #351.

However, I have to update new unit tests that I added for this feature as some of them are not passing in earlier node versions in travis. Basically I shouldn't have used fs mode. So, I am changing the new tests to use `fs.stat.mtime.getTime()` instead of `fs.stat.mode` for filtering dest. I will update the PR right away then.

**Edit**

I fixed the tests by using `fs.stat` `birthtime` instead of `mtime`. Now all tests passed!

### DIFF
--- a/docs/copy.md
+++ b/docs/copy.md
@@ -26,3 +26,23 @@ fs.copy('/tmp/mydir', '/tmp/mynewdir', function (err) {
   console.log('success!')
 }) // copies directory, even if it has subdirectories or files
 ```
+
+**Using filter function**
+
+```js
+var fs = require('fs-extra')
+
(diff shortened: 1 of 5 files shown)


manidlou:
I will squash this then.

manidlou:
I close this and open a new clean PR.

Issue #358, opened 2017-02-16
fs.copy silently fails on windows


arekkas:
This is an upstream issue report from [electron-packager#544](https://github.com/electron-userland/electron-packager/issues/544) and in particular [this comment](https://github.com/electron-userland/electron-packager/issues/544#issuecomment-280419995). The issue exists in both 1.0 and 2.0 versions.

electron-packager is a cli helper that packages an electron app. It does so by copying a node app (including node_modules) to the temp directory. This step uses `fs.copy` which silently fails. Neither the callback is called, nor an error is thrown.

This is the line which fails silently

```js
fs.copy(opts.dir, appPath, {filter: ignore.userIgnoreFilter(opts), dereference: shouldDeref}, cb)
```

I'd be glad to help you identify root cause of this issue. My current guess that it's combination of filter/dereference and maybe (?) long paths?

Note that this *only* happens on windows. On OSX, packaging the project works fine.

jprichardson:
@arekkas thank you for your report and offer to help.

It seems that `fs.copy()` all of a sudden is failing for a large swath of people. See in particular:

- https://github.com/jprichardson/node-fs-extra/issues/357
- https://github.com/jprichardson/node-fs-extra/issues/326
- https://github.com/jprichardson/node-fs-extra/issues/351#issuecomment-278752763

Let's start with this... which version of Node are you using? Does changing the version of Node to a previous version (in particular, older major), still cause the issue?

arekkas:
Thanks for your swift reply. I'm currently using node 6.6.0. Unfortunately, there is no `nvm` I am aware of that works on windows. So up-/downgrading is a bit of a pain. I'd do it however if you're confident that that might be the issue.

jdalton:
> Unfortunately, there is no nvm I am aware of that works on windows

I use `nvm` with BashOnWindows in Windows 10.

jprichardson:
@arekkas https://github.com/coreybutler/nvm-windows However, please only switch nvm windows if you think you'll get personal value out of it. Thanks for your willingness to help here.

arekkas:
Ok, installed. Tried majors 4, 5, 6. All of them have (apparently) the same issue.

arekkas:
I did some good old `console.log` debugging. @jprichardson it looks like the culprit is the counter in `ncp.js`:

```
  function doneOne (skipped) {
    console.log('doneOne 1', skipped)
    if (!skipped) running--
[...]

PR #366, opened 2017-02-22
Add dest parameter to filter option in copy and copySync

### LINKED ISSUE #351: Additional param for options.filter() on copy()

manidlou:
This resolves #351.

I added dest as an additional parameter to filter option in both `copy` and `copySync`.

### DIFF
--- a/lib/copy-sync/__tests__/copy-sync-dir.test.js
+++ b/lib/copy-sync/__tests__/copy-sync-dir.test.js
@@ -64,7 +64,32 @@ describe('+ copySync()', () => {
       assert.strictEqual(link, 'destination')
     })
 
-    it('should should apply filter recursively', () => {
+    describe('> when the destination dir does not exist', () => {
+      it('should create the destination directory and copy the file', () => {
+        const src = path.join(TEST_DIR, 'data/')
+        fs.mkdirSync(src)
+
+        const d1 = 'file1'
+        const d2 = 'file2'
+
+        fs.writeFileSync(path.join(src, 'f1.txt'), d1)
+        fs.writeFileSync(path.join(src, 'f2.txt'), d2)
+
+        const dest = path.join(TEST_DIR, 'this/path/does/not/exist/outputDir')
+
+        fs.copySync(src, dest)
+
+        const o1 = fs.readFileSync(path.join(dest, 'f1.txt'), 'utf8')
+        const o2 = fs.readFileSync(path.join(dest, 'f2.txt'), 'utf8')
+
+        assert.strictEqual(d1, o1)
+        assert.strictEqual(d2, o2)
+      })
+    })
+  })
+
+  describe('> when filter is used', () => {
+    it('should should apply filter recursively', done => {
       const FILES = 2
       // Don't match anything that ends with a digit higher than 0:
       const filter = s => /(0|\D)$/i.test(s)
@@ -104,9 +129,10 @@ describe('+ copySync()', () => {
           assert(!fs.existsSync(path.join(destSub, j.toString())))
         }
       }
+      done()
     })
 
-    it('should apply the filter to directory names', () => {
+    it('should apply the filter to directory names', done => {
       const IGNORE = 'ignore'
       const filter = p => !~p.indexOf(IGNORE)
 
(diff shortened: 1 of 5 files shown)


lib/copy-sync/__tests__/copy-sync-dir.test.js
```diff
@@ -121,29 +147,61 @@ describe('+ copySync()', () => {
 
       assert(!fs.existsSync(path.join(dest, IGNORE)), 'directory was not ignored')
       assert(!fs.existsSync(path.join(dest, IGNORE, 'file')), 'file was not ignored')
+      done()
```
RyanZim: Revert this change. Using `done` isn't needed since we are using sync methods here.

lib/copy-sync/__tests__/copy-sync-dir.test.js
```diff
@@ -104,9 +129,10 @@ describe('+ copySync()', () => {
           assert(!fs.existsSync(path.join(destSub, j.toString())))
         }
       }
+      done()
```
RyanZim: Revert this change. Using `done` isn't needed since we are using sync methods here.

lib/copy-sync/copy-sync.js
```diff
@@ -35,8 +35,8 @@ function copySync (src, dest, options) {
 
   if (options.filter instanceof RegExp) {
     console.warn('Warning: fs-extra: Passing a RegExp filter is deprecated, use a function')
-    performCopy = options.filter.test(src)
-  } else if (typeof options.filter === 'function') performCopy = options.filter(src)
+    performCopy = options.filter.test(src, dest)
```
RyanZim: This change doesn't do anything; `Regex.prototype.test` only accepts one param.

lib/copy/ncp.js
```diff
@@ -37,11 +37,11 @@ function ncp (source, dest, options, callback) {
     if (filter) {
       if (filter instanceof RegExp) {
         console.warn('Warning: fs-extra: Passing a RegExp filter is deprecated, use a function')
-        if (!filter.test(source)) {
+        if (!filter.test(source, dest)) {
```
RyanZim: This change doesn't do anything; `Regex.prototype.test` only accepts one param.

lib/copy-sync/copy-sync.js
manidlou: Very true. I missed it. Thanks @RyanZim. Great that you caught this.

manidlou:
I fixed them. Please let me know if there is any other issues left.

lib/copy-sync/__tests__/copy-sync-dir.test.js
```diff
@@ -147,50 +146,47 @@ describe('+ copySync()', () => {
...
+        fs.mkdirsSync(src)
+        fs.writeFileSync(path.join(src, 'somefile.html'), 'some data')
         fs.mkdirsSync(dest)
         try {
           fs.copySync(src, dest, filter)
         } catch (err) {
           assert.ifError(err)
         }
-        assert(!fs.existsSync(path.join(dest, 'subdir')))
-        done()
```
RyanZim: You still need `done` when you're using setTimeout (since that makes the code async).

manidlou:
Should I merge this?

RyanZim:
@manidlou It's up to you. I was leaving it for @jprichardson, since I'm a bit tired today and could have missed something.

---

@jprichardson Perhaps we should formalize some collaborator guidelines sometime?

lib/copy-sync/__tests__/copy-sync-dir.test.js
```diff
@@ -27,15 +27,15 @@ describe('+ copySync()', () => {
       src = path.join(TEST_DIR, 'src')
       dest = path.join(TEST_DIR, 'dest')
 
-      fs.mkdirsSync(src)
+      fs.mkdirSync(src)
```
jprichardson: Why the conversion here to `mkdirSync` from `mkdirsSync`?

RyanZim: I can't speak for @manidlou, but `mkdirs` technically isn't needed, since we know the parent directory exists.

manidlou: > @jprichardson Perhaps we should formalize some collaborator guidelines sometime? :+1: 

> Why the conversion here to mkdirSync from mkdirsSync?

As @RyanZim mentioned, I thought we wouldn't need to call `mkdirs` since we know the parent directory exist. So, I guess I should revert to `mkdirsSync`. Is that right?

And also sorry that I am still not in harmony with you guys. I try my best to be as helpful as possible.

jprichardson: > And also sorry that I am still not in harmony with you guys. I try my best to be as helpful as possible.

@manidlou you're very helpful, thank you!!

> @jprichardson Perhaps we should formalize some collaborator guidelines sometime?

I think for now, any modifications (including my own), barring some crazy necessary hot-fix, require at least one other reviewer? 

To be clear, it's okay to accept a self-submitted PR, but it must be for a very good reason. We're all professionals here, so I think this loose guideline should be acceptable for now.

Sound good @manidlou @RyanZim?

jprichardson:
Let me know if there is anything else that's needed before I publish 2.1.0.

manidlou:
Sounds good.

Just a quick question, for issues like #358 that user doesn't experience the issue anymore, is this a correct action to comment something like "closing the issue" and then closing it?

By the way, I close this branch as it is already merged.

RyanZim:
> And also sorry that I am still not in harmony with you guys. I try my best to be as helpful as possible.

You're doing great.

> Just a quick question, for issues like #358 that user doesn't experience the issue anymore, is this a correct action to comment something like "closing the issue" and then closing it?

Generally, if the user can't reproduce, then yes, you would comment & close. #358 is a gray area, since it is a problem on other Node versions. I'm personally not sure what to do here. @jprichardson?

PR #367, opened 2017-02-22
Refactor tests to ES6

JPeer264:
(#355)
In commit https://github.com/jprichardson/node-fs-extra/commit/55aadcf1f1871200535c349e82e33c4b55c3337f I removed `semver`, as dependency as it was just used to ignore Node v0.10.

### DIFF
--- a/lib/json/__tests__/output-json.test.js
+++ b/lib/json/__tests__/output-json.test.js
@@ -1,49 +1,53 @@
-var assert = require('assert')
-var fs = require('fs')
-var os = require('os')
-var path = require('path')
-var fse = require(process.cwd())
+'use strict'
+
+const fs = require('fs')
+const os = require('os')
+const fse = require(process.cwd())
+const path = require('path')
+const assert = require('assert')
+const outputJson = require('../output-json')
 
 /* global beforeEach, describe, it */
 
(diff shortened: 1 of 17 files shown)


lib/json/__tests__/output-json.test.js
```diff
@@ -1,49 +1,53 @@
...
+      it('should pass the option along to jsonfile module', done => {
+        const file = path.join(TEST_DIR, 'this-dir', 'does-not', 'exist', 'really', 'file.json')
         assert(!fs.existsSync(file))
 
-        var replacer = function (k, v) { if (v === 'JP') return 'Jon Paul'; else return v }
-        var data = {name: 'JP'}
+        const replacer = (k, v) => {
+          if (v === 'JP') return 'Jon Paul'
+          return v
+        }
```
RyanZim: Could be shortened to:

    const replacer = (k, v) => v === 'JP' ? 'Jon Paul' : v

It's up to you if you want to shorten this or not.

JPeer264: That's a good one - it is updated. Thanks!

jprichardson:
Thanks @JPeer264!

Issue #83, opened 2014-09-09
Prevent copying a directory into itself


iwege:
`fse.copy('src','src/dest')`  will create the dest  continues to recurse, and finally it throws an error `ENAMETOOLONG`

It links to AvianFlu/ncp#4

jprichardson:
Off the cuff, I'd say that I could just resolve the paths and check if one is a substring of the other. But I'd be concerned about introducing regressions. Thoughts?

iwege:
I have collected some cases in this pr https://github.com/AvianFlu/ncp/pull/61, maybe you can use it as a reference.

jprichardson:
@lwege are you install interested in this getting fixed? I'd like to tackle this - it'd be awesome to take your list of cases and turn them into tests for both *nix and Windows.

iwege:
@jimhigson I copied testcase to  https://github.com/iwege/node-fs-extra/tree/feature/prevent-copy-to-self , but how can I test it first? Or do I need to fix this issue with my old code?

jprichardson:
Would you link directly to the test case so that I can take a look? Thanks.

iwege:
@jprichardson https://github.com/iwege/node-fs-extra/commit/0dc1610ff98ad5a556ce3956d229a7472bf8fb01

jprichardson:
Thank you. Tests look good. Any thoughts on a reliable fix?

iwege:
I have used the https://github.com/AvianFlu/ncp/pull/61 in my project and no one reports relative issue to me. But I don't know it reliable or not.

DaneEveritt:
Sorry to bring this back up, but any progress on implementing this, or should I add my own checks?

jprichardson:
> or should I add my own checks

What checks would you add?

DaneEveritt:
I built off @iwege's changes, and made the following function that I'm using before I call the move function.

```javascript
isSelf(moveTo, moveFrom) {
    const target = this.server.path(moveTo);
    const source = this.server.path(moveFrom);

    if (!_.startsWith(target, source)) {
        return false;
    }

    const end = target.slice(source.length);
    if (!end) {
        return true;
    }

    return _.startsWith(end, '/');
}
```

Prevents moving folders into themselves, but doesn't block moving a folder into another folder that *starts* with the same name (so `src` can still move into `src-dest`, but not into `src`).

Renaming files to the same name also triggers the catch, but that doesn't bother me so much.

[...]

PR #374, opened 2017-02-26
[WIP] Rewrite lib/copy/ncp.js

### LINKED ISSUE #198: handle copying between identical files
I hope this module can be as clever as `cp`. It should recognize copying between identical files and do nothing.

Now, v0.26.2, neither `copy` nor `copySync` can handle these cases correctly.
- If `src` and `dst` are the same file
  - It opens a read stream first, and then a write stream, which results in the file becomes empty.
- If `src` is symbolic link and points to `dst`
  - By default, option `clobber` is `true`, so that `dst` will be removed, which means the real file is removed. If `clobber` is `false`, the file becomes empty.
- If `dst` is symbolic link and points to `src`
  - Similar to the previous case, `dst` is removed and a new file is created, then the link is not existed any more. Or the file becomes empty.

I have added corresponding test cases in [my branch](https://github.com/chinesedfan/node-fs-extra/tree/copy_identical).

manidlou:
Refs: #292.

This is the first draft of reconsidering/rewriting `lib/copy/ncp.js`.

As @RyanZim mentioned before:

> Once rewritten, it may be small enough to inline in lib/copy/copy.js;

It make sense to me too. So, it is wrapped in one file now `lib/copy/copy.js`.

A few notes:

- In `checkDest()`, the reason that I directly use `fs.readlink()` as apposed to getting the `fs.stat()` then decide accordingly, is to avoid possibility of race since we essentially need to check src against a possible resolved dest path. See #198.

- I added more tests for symlinks, copying identical, and copying into itself as well. These changes should resolve #198 and #83.

- I activated the `move()` skipped tests: https://github.com/jprichardson/node-fs-extra/blob/master/lib/move/__tests__/move.test.js#L310. It passes now.

- Currently all tests pass including old and new tests.

- The reason that I didn't create the branch here is I figured that'd be better to keep the upstream repo as clean as possible since this one probably take some time to discuss and decide on and whenever we want we can migrate it here.

Please feel free to point out any issues that you can think of with these changes.


manidlou:
Hmm! coverage decreased! I will look at it and try to add tests for the parts that are not covered yet.

lib/copy/__tests__/copy-prevent-copying-into-itself.test.js
```diff
@@ -0,0 +1,445 @@
...
+    src = path.join(TEST_DIR, 'src')
+    fs.mkdirsSync(src)
+
+    // ensureFileSync creates required parent dirs for us:)
+    FILES.forEach(f => fs.ensureFileSync(path.join(src, f)))
+
+    fs.writeFileSync(path.join(src, FILES[0]), dat0)
+    fs.writeFileSync(path.join(src, FILES[1]), dat1)
+    fs.writeFileSync(path.join(src, FILES[2]), dat2)
+    fs.writeFileSync(path.join(src, FILES[3]), dat3)
```
RyanZim: Could use `outputFile` here instead of `ensureFile` and `writeFile`.

lib/copy/__tests__/copy-prevent-copying-into-itself.test.js
```diff
@@ -0,0 +1,445 @@
...
+
+        dest = path.join(TEST_DIR, 'src_src', 'dest')
+        testSuccess(srcLink, dest, () => {
+          const link = fs.readlinkSync(dest)
+          assert.strictEqual(link, src)
+          done()
+        })
+      })
+
+      it(`should copy the directory successfully when dest is 'srcsrc/dest'`, done => {
```
RyanZim: According to the test's code, the description should say: `when dest is 'src_src/dest'` (missing underscore).

lib/copy/__tests__/copy.test.js
```diff
@@ -159,6 +159,7 @@ describe('fs-extra', () => {
         })
       })
 
+      /* It is redundant. See above, we already tested for this case.
```
RyanZim: Just remove this.

lib/copy/__tests__/copy.test.js
```diff
@@ -196,6 +198,19 @@ describe('fs-extra', () => {
...
       })
 
+      it('should not copy and return when nothing matched', done => {
+        const srcFile1 = path.join(TEST_DIR, '1.jade')
+        fs.writeFileSync(srcFile1, '')
+        const destFile1 = path.join(TEST_DIR, 'dest1.jade')
+        const filter = /.html$|.css$/i
+
+        fse.copy(srcFile1, destFile1, filter, (err) => {
+          assert(!err)
```
RyanZim: Could use `assert.ifError(err)` here.

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+  options.overwrite = 'overwrite' in options ? !!options.overwrite : options.clobber
+  options.dereference = 'dereference' in options ? !!options.dereference : false
+  options.preserveTimestamps = 'preserveTimestamps' in options ? !!options.preserveTimestamps : false
+
+  options.filter = options.filter || function () { return true }
+
+  // Warn about using preserveTimestamps on 32-bit node
   if (options.preserveTimestamps && process.arch === 'ia32') {
     console.warn(`fs-extra: Using the preserveTimestamps option in 32-bit node is not recommended;\n
     see https://github.com/jprichardson/node-fs-extra/issues/269`)
```
RyanZim: Could port this to ES6 template strings.

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+      console.warn('Warning: fs-extra: Passing a RegExp filter is deprecated, use a function')
+      if (!options.filter.test(src)) return cb()
+    } else if (typeof options.filter === 'function') {
+      if (!options.filter(src, dest)) return cb()
+    }
+  }
+
+  const destParent = path.dirname(dest)
+  mkdirs(destParent, err => {
+    if (err && err.code !== 'EEXIST') return cb(err)
```
RyanZim: `mkdirs` doesn't return `EEXIST` errors, so no need for that logic. Just do:

    if (err) return cb(err)

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+}
+
+function checkDest (dest, cb) {
+  fs.readlink(dest, (err, resolvedDestPath) => {
+    if (err) {
+      if (err.code === 'ENOENT') return cb(null, DEST_NOENT)
+      // dest exists but is not a link
+      else if (err.code === 'EINVAL' || err.code === 'UNKNOWN') return cb(null, DEST_EXISTS)
+      else return cb(err)
+    } else { // dest exists and is a link
```
RyanZim: No need for the `else` here, since you are returning in the `if` statement.

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
-const ncp = require('./ncp')
-const mkdir = require('../mkdirs')
+const mkdirs = require('../mkdirs').mkdirs
+const utimes = require('../util/utimes').utimesMillis
 
-function copy (src, dest, options, callback) {
-  if (typeof options === 'function' && !callback) {
-    callback = options
+const DEST_NOENT = -1
+const DEST_EXISTS = 1
```
RyanZim: Possibly could/should use ES6 `Symbol`s here?

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+  ws.on('error', err => cb(err))
+
+  ws.on('open', () => {
+    rs.pipe(ws)
+  }).once('close', () => {
+    fs.chmod(dest, srcStat.mode, err => {
+      if (err) return cb(err)
+      if (options.preserveTimestamps) {
+        return utimes(dest, srcStat.atime, srcStat.mtime, cb)
+      } else return cb()
```
RyanZim: Don't need `else` here.

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+        })
       })
+    } else if (res === DEST_EXISTS) {
+      if (isSrcSubdir(src, dest)) return cb(new Error(`Cannot copy directory '${src}' into itself '${dest}'`))
+      return cpDir(src, dest, options, cb)
+    } else if (res && typeof res !== 'number') { // dest exists and is a link
+      if (isSrcSubdir(src, res)) return cb(new Error(`Cannot copy directory '${src}' into itself '${res}'`))
+      if (src === res) return cb()
+      return cpDir(src, dest, options, cb)
+    } else return cb()
```
RyanZim: In what case would this code path be taken?

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+        copy(path.join(src, item), path.join(dest, item), options, err => {
+          if (err) reject(err)
+          else resolve()
+        })
+      })
+    })).then(() => {
+      return cb()
+    }).catch(err => {
+      return cb(err)
+    })
```
RyanZim: Should be shortened to
```js
})).then(() => cb())
.catch(cb)
```

lib/copy/copy.js
```diff
@@ -2,51 +2,213 @@
...
+          return cb(new Error(`Cannot copy directory '${resolvedSrcPath}' into itself '${dest}'`))
+        }
+        return fs.symlink(resolvedSrcPath, dest, cb)
+      } else if (resolvedDestPath === DEST_EXISTS) { // dest exists but is not a link
+        if (isSrcSubdir(resolvedSrcPath, dest)) {
+          return cb(new Error(`Cannot copy directory '${resolvedSrcPath}' into itself '${dest}'`))
+        }
+        // if src points to dest
+        if (resolvedSrcPath === dest) return cb()
+        return copy(resolvedSrcPath, dest, options, cb)
```
RyanZim: I'm a bit hairy on expected behavior for symlinks; @jprichardson is this correct?

lib/copy/__tests__/copy-prevent-copying-into-itself.test.js
manidlou: That's right. In fact, description is correct, but dest path in test's code should change to match the description.

lib/copy/copy.js
manidlou: I like the idea of using `Symbol`s here as one of their use cases is for unique constants.

lib/copy/copy.js
manidlou: @RyanZim would you please say what you mean exactly? is it referring to the extra `\n` in the warn?

lib/copy/copy.js
manidlou: Honestly, I am not 100% sure neither. So, @jprichardson we definitely need some advice here.

lib/copy/copy.js
manidlou: If you mean line 141 `} else return cb()`, I guess it is not needed and should be removed as we check for all possible conditions.

lib/copy/copy.js
RyanZim: Oops, nevermind.

manidlou: Is that `\n` necessary as with template strings we can do multi-line like normally?

RyanZim: @manidlou We are already using template strings (I didn't notice that when I commented here). The `\n` is to make a double linebreak.

Just ignore this line comment.

lib/copy/copy.js
```diff
@@ -2,51 +2,211 @@
...
+        if (err) return cb(err)
+        fs.chmod(dest, srcStat.mode, err => {
+          if (err) return cb(err)
+          return cpDir(src, dest, options, cb)
+        })
+      })
+    } else if (res === DEST_EXISTS) {
+      if (isSrcSubdir(src, dest)) return cb(new Error(`Cannot copy directory '${src}' into itself '${dest}'`))
+      return cpDir(src, dest, options, cb)
+    } else if (res && typeof res !== 'number') { // dest exists and is a link
```
RyanZim: This `typeof res !== 'number'` is obsolete since you are using symbols.

lib/copy/copy.js
```diff
@@ -2,51 +2,211 @@
...
+        }
+        return fs.symlink(resolvedSrcPath, dest, cb)
+      } else if (resolvedDestPath === DEST_EXISTS) { // dest exists but is not a link
+        if (isSrcSubdir(resolvedSrcPath, dest)) {
+          return cb(new Error(`Cannot copy directory '${resolvedSrcPath}' into itself '${dest}'`))
+        }
+        // if src points to dest
+        if (resolvedSrcPath === dest) return cb()
+        return copy(resolvedSrcPath, dest, options, cb)
+      } else if (resolvedDestPath && typeof resolvedDestPath !== 'number') { // dest is a link
```
RyanZim: Same here.

lib/copy/copy.js
RyanZim: @jprichardson ping?

lib/copy/copy.js
```diff
@@ -2,51 +2,211 @@
...
+    } else if (res === DEST_EXISTS) {
+      if (options.overwrite) {
+        fs.unlink(dest, err => {
+          if (err) return cb(err)
+          return cpFile(src, srcStat, dest, options, cb)
+        })
+      } else if (options.errorOnExist) {
+        return cb(new Error(dest + ' already exists'))
+      } else return cb()
+    } else return cb()
```
manidlou: This last `else` basically means dest exists and is a symlink. I believe the same pattern that is used for the case `DEST_EXISTS`, which means checking for `options.overwrite` and operates accordingly, should be applied for this case as well. Is that right?

lib/copy/copy.js
manidlou: Also, I just noticed that for both cases where dest exists (regular and symlink), `onDir()` should check for `options.overwrite` and proceed its operation accordingly; the same as `onFile()`.

lib/copy/copy.js
manidlou: Again here also, when dest exists (regular and symlink), should go with the same pattern of checking `options.overwrite` and operates accordingly.

lib/copy/__tests__/copy-prevent-copying-identical.test.js
```diff
@@ -0,0 +1,192 @@
...
+      })
+    })
+
+    describe(`>> when src is a symlink that points to a regular dest`, () => {
+      it('should not copy and return', done => {
+        dest = path.join(TEST_DIR, 'dest')
+        fs.mkdirsSync(dest)
+        const subdir = path.join(TEST_DIR, 'dest', 'subdir')
+        fs.mkdirsSync(subdir)
+        fs.writeFileSync(path.join(subdir, 'file.txt'), 'some data')
```
manidlou: Will be changed to use `fs.outputFileSync()`.

lib/copy/__tests__/copy-prevent-copying-identical.test.js
```diff
@@ -0,0 +1,192 @@
...
+  //  src is regular, dest is symlink
+  //  src is symlink, dest is regular
+  //  src is symlink, dest is symlink
+
+  describe('> when the source is a file', () => {
+    describe(`>> when src is regular and dest is a symlink that points to src`, () => {
+      it('should not copy and return', done => {
+        src = path.join(TEST_DIR, 'src', 'somefile.txt')
+        fs.ensureFileSync(src)
+        fs.writeFileSync(src, 'some data')
```
manidlou: Same here. It will be changed to use `fs.outputFileSync()` instead.

lib/copy/__tests__/copy-prevent-copying-identical.test.js
```diff
@@ -0,0 +1,192 @@
...
+          done()
+        })
+      })
+    })
+
+    describe('>> when src and dest are symlinks that point to the exact same path', () => {
+      it('should not copy and return', done => {
+        src = path.join(TEST_DIR, 'src', 'srcfile.txt')
+        fs.ensureFileSync(src)
+        fs.writeFileSync(src, 'src data')
```
manidlou: Same here.

lib/copy/__tests__/copy-prevent-copying-identical.test.js
```diff
@@ -0,0 +1,192 @@
...
+          done()
+        })
+      })
+    })
+
+    describe(`>> when src is a symlink that points to a regular dest`, () => {
+      it('should not copy and return', done => {
+        dest = path.join(TEST_DIR, 'dest', 'somefile.txt')
+        fs.ensureFileSync(dest)
+        fs.writeFileSync(dest, 'some data')
```
manidlou: Same here.

lib/copy/copy.js
RyanZim: @manidlou Historically, `overwrite` only is checked for files. There's no need to nuke an entire directory if we just need to add one file to it. I'd leave it that way.

lib/copy/copy.js
RyanZim: I'm not sure. @jprichardson ?

jprichardson: Agreed @manidlou

lib/copy/copy.js
manidlou: @RyanZim I guess you are right. I don't have a good feeling about `return copy(resolvedSrcPath, dest, options, cb)` neither. I recall I used it because of the ability to handle weird cases like when src is a symlink and dest exists (either regular or another symlink) and is a kid of src as well.

`return fs.symlink(resolvedSrcPath, dest, cb)` correctly throws `EEXIST: file already exists, symlink 'src' -> 'dest'` for those cases.

Now, I think we should decide what we want to do here for these two cases when src is a symlink:

- dest exists and is a regular file/directory
- dest exists and is a symlink (file/directory)

Do we want to throw an error and notify user that dest exists?

RyanZim:
@manidlou ping?

manidlou:
@RyanZim I asked a question a few comments above regarding copying when src is a symlink, the case that we are not sure about the correct behavior yet, that I've been waiting for your and @jprichardson's answers. Would you please take a look at it? :smiley:

lib/copy/copy.js
jprichardson: > Now, I think we should decide what we want to do here for these two cases when src is a symlink:

What is the existing behavior?

manidlou: > What is the existing behavior?

Currently, when src is a symlink, this is what happens for the following cases:

- dest exists and is a regular file or directory

  it throws `Error: EINVAL: invalid argument, readlink 'dest'`, and this is because the existing behavior assumes when src is a symlink, dest is also a symlink (https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy/ncp.js#L182).

- dest exists and is a symlink

  since `overwrite` is the default behavior, it removes dest then copies src (creates a symlink in dest that points to src).

Now, for the case 2, it works fine except if src is a kid of dest. In this case, since it removes dest, we end up losing src and consequently a broken symlink is created.

manidlou:
So, following our discussion on copying symlinks when src is a dest's kid and vice versa, to me for this case (which I guess is relatively a rare case) it is reasonable to throw error and let user decide on the action. Although we can deliberately handle these cases, but that makes it overly complex while it think it is not worth it for this case. I tried to find a short and yet descriptive error message for this case. I came up with `${dest} exists and contains ${src}`.

@jprichardson @RyanZim what do you think then?

manidlou:
What about `noRecurseOnFailedFilter`? do you think it is reasonable enough to keep it or it's better to remove it for the sake of simplicity?

RyanZim:
@jprichardson Thoughts on this?

manidlou:
@jprichardson @RyanZim will you please take a look at https://github.com/manidlou/benchmark-fs-extra-copy?

I add a quick summary here too:

I've been working on an alternative implementation for `copy()` that uses `recursive` function (I couldn't find a better name) instead of `Promise` for copying a directory to see if we can get a faster execution. In addition, I organized the code into bunch of small functions as they can be optimized by v8 particularly when copying directories with lots of files and subdirectories as the same function needs to be called repeatedly. You've probably seen these posts or alike before: [A tour of V8: Crankshaft, the optimizing compiler](http://jayconrod.com/posts/54/a-tour-of-v8-crankshaft-the-optimizing-compiler) and [NodeJS : A quick optimization advice](https://top.fse.guru/nodejs-a-quick-optimization-advice-7353b820c92e).

So, I created a benchmark to compare the running time of these implementations:

- this new `recursive` function
- `Promise`
- `current` (existing)

to see which one runs faster particularly on large input sizes, when a src directory has lots of items.

So, when src length is up to `~6000`, they run pretty much the same. However, on roughly `> 6000` the new `recursive` function runs significantly faster than other implementations.

[line chart comparison, range: 110-8190](https://s3-us-west-2.amazonaws.com/benchmark-fs-extra-copy/recursive-promise-current_half.html)

[line chart comparison, range: 110-111110](https://s3-us-west-2.amazonaws.com/benchmark-fs-extra-copy/recursive-promise-current.html)

any thoughts please?

RyanZim:
@manidlou Sorry so terribly long in getting back (again). If you're comfortable with the `recursive` method; by all means add it. Just glancing at it, it doesn't look bad, and better perf is always good.

RyanZim:
BTW @manidlou, please don't let me just let stuff slide like this; if I don't reply within 7 days, you're welcome to ping me (I forget stuff really easily; if I'm not responding, I could be busy, but most of the time, I just forgot about it).

manidlou:
@RyanZim besides perf, overall I personally have a better feeling about `recursive` than `Promise`. So, if you don't mind, since it's been a long time that this PR been open, just for saving a bit of time resolving conflicts manually, I'll open a new PR for `recursive` implementation and close this one. Of course, if you think we need to keep this open, please don't hesitate to reopen it! :smile:

PR #375, opened 2017-02-27
Refactor test.js

JPeer264:
I totally forgot one file to refactor 😅 

`.on('end', () => ...)` can be written as one liner, but there is no real benefit, as it is harder to read.

Thanks again for this awesome library and that I could help you out 👍


jprichardson:
Thanks again @JPeer264!!

PR #376, opened 2017-03-02
move(): fall back to streams when hard links aren't supported

tvald:
`move()` should fall back to using streams when `fs.link()` fails with `ENOTSUP`, just as it already does for `EPERM` and other errors.

(As discussed in nodejs/node#11663, hard links aren't supported on ReFS.)


RyanZim:
@tvald Thanks!

Issue #377, opened 2017-03-03
fs.move should noop when source and desc are the same.


jdalton:
Since the file is already at its destination `fs.move` should noop. Similar to how `fs.remove` doesn't error if a file doesn't exist.

RyanZim:
@jdalton I agree this case should be handled. Interested in submitting a PR?

BTW, what's the current behavior?

jdalton:
Yep. I'll submit a PR. The current behavior is to throw an error (some already-exists error)

jdalton:
Moved to #378.

PR #378, opened 2017-03-03
Don't move when source and dest paths are the same.

jdalton:
This PR addresses #377 to avoid a move operation if the `source` and `dest` paths are the same.

*(I'll add tests too, just wanted to get the ball rolling)*

### DIFF
--- a/lib/move/index.js
+++ b/lib/move/index.js
@@ -35,7 +35,9 @@ function move (source, dest, options, callback) {
   }
 
   function doRename () {
-    if (overwrite) {
+    if (path.resolve(source) === path.resolve(dest)) {
+      setImmediate(callback)
+    } else if (overwrite) {
       fs.rename(source, dest, err => {
         if (!err) return callback()
 
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -69,6 +69,26 @@ describe('move', () => {
     })
   })
 
+  it('should not move a file if source and destination are the same', done => {
+    const src = `${TEST_DIR}/a-file`
+    const dest = src
+
+    fse.move(src, dest, err => {
+      assert.ifError(err)
+      done()
+    })
+  })
+
+  it('should not move a directory if source and destination are the same', done => {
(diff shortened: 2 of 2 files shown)


lib/move/index.js
```diff
@@ -35,7 +35,9 @@ function move (source, dest, options, callback) {
   }
 
   function doRename () {
-    if (overwrite) {
+    if (path.resolve(source) === path.resolve(dest)) {
+      setTimeout(callback, 1)
+    } else if (overwrite) {
```
jdalton: `fs.move` is async do you want it to exit early as such too (the `setTimeout`).
Do y'all have a pattern in place for this already?

(I'll add tests too, just wanted to get the ball rolling)

jprichardson: @jdalton what are your thoughts on `setTimeout` over `setImmediate`? It's my understanding `setTimeout` with an input of 1, still clamps to 15 ms.

jdalton: > what are your thoughts on setTimeout over setImmediate? It's my understanding setTimeout with an input of 1, still clamps to 15 ms.

I have no strong opinions on it here since the primary goal is to just defer it a bit. Based on that goal [setImmediate](https://nodejs.org/api/timers.html#timers_setimmediate_callback_args) looks like a better call.

jprichardson:
Before the behavior would error out, right? Should we consider this breaking? Thoughts @jdalton?

jdalton:
It did error out. I would consider this a bug fix. *(I haven't forgotten about the unit tests, just busy)*

manidlou:
What should be the correct behavior then? Since whatever that'd be, the same behavior should be applied to `moveSync` as well.

jdalton:
>What should be the correct behavior then? Since whatever that'd be, the same behavior should be applied to `moveSync` as well.

I'm asserting by the title and behavior of this PR "Don't move when source and dest paths are the same."

jdalton:
Added tests.

jprichardson:
Thank you @jdalton!

PR #379, opened 2017-03-05
Document `outputFile()` only supports a file path

hail2u:
`fs.writeFile()` supports a buffer or a file descriptor as 1st  argument (`file`), but `fs-extra.outputFile()` does not. Because it sends `file` parameter to `path.dirname()` that requires `String`.


RyanZim:
@hail2u Thanks!

PR #381, opened 2017-03-07
Add moveSync and its tests

### LINKED ISSUE #309: moveSync support
Would you like to have `moveSync` just like `copy` and `copySync`? Thanks :smile:

manidlou:
This should resolves #309.

Manually tested on Ubuntu 16.04 and Windows 10 on virtualbox moved across drives succesfully, cases such as `moveSync('G:\\src', 'E:\\nested\\dest')`.

Like always, please feel free to point out any issues that you can think of :smiley:.


lib/move-sync/__tests__/move-sync.test.js
```diff
@@ -0,0 +1,411 @@
...
+  afterEach(done => rimraf(TEST_DIR, done))
+
+  it('should not move if src and dest are the same', () => {
+    const src = `${TEST_DIR}/a-file`
+    const dest = `${TEST_DIR}/a-file`
+
+    try {
+      fse.moveSync(src, dest)
+    } catch (err) {
+      assert.ifError(err)
```
RyanZim: All these `try..catch` blocks are unnecessary; if `moveSync` fails, it will automatically fail the test with the error message. Adding a `try..catch` with an `assert.ifError` is just redundant.

lib/move-sync/__tests__/move-sync.test.js
```diff
@@ -0,0 +1,411 @@
...
+
+  // verify dest directory does not exist
+    assert(!fs.existsSync(path.dirname(dest)))
+
+    fse.move(src, dest, {mkdirp: false}, err => {
+      assert.strictEqual(err.code, 'ENOENT')
+      done()
+    })
+  })
+  */
```
RyanZim: Remove this test.

lib/move-sync/__tests__/move-sync.test.js
```diff
@@ -0,0 +1,411 @@
...
+    try {
+      fse.moveSync(src, dest)
+    } catch (err) {
+      assert.ifError(err)
+    }
+
+    const contents = fs.readFileSync(dest, 'utf8')
+    const expected = /^sonic the hedgehog\r?\n$/
+    assert.ok(contents.match(expected), `${contents} match ${expected}`)
+    tearDownMockFs('EXDEV')
```
RyanZim: No need to pass a parameter here, `tearDownMockFs` doesn't use it. Correct other places where this is done.

lib/move-sync/__tests__/move-sync.test.js
```diff
@@ -0,0 +1,411 @@
...
+        assert.ifError(err)
+      }
+
+      const contents = fs.readFileSync(dest, 'utf8')
+      const expected = /^sonic the hedgehog\r?\n$/
+      assert.ok(contents.match(expected), `${contents} match ${expected}`)
+    })
+  })
+
+  describe.skip('> when trying to a move a folder into itself', () => {
```
RyanZim: Why the `.skip`?

lib/move-sync/__tests__/move-sync.test.js
```diff
@@ -0,0 +1,411 @@
...
+      } catch (err) {
+        assert(err)
+        assert(fs.existsSync(SRC_DIR))
+      }
+    })
+  })
+
+  // tested on Linux ubuntu 3.13.0-32-generic #57-Ubuntu SMP i686 i686 GNU/Linux
+  // this won't trigger a bug on Mac OS X Yosimite with a USB drive (/Volumes)
+  // see issue #108
```
RyanZim: Does this comment apply to `moveSync`?

lib/move-sync/index.js
```diff
@@ -0,0 +1,116 @@
+'use strict'
+
+// most of this code was written by Andrew Kelley
+// licensed under the BSD license: see
+// https://github.com/andrewrk/node-mv/blob/master/package.json
+// This is the sync version that somehow follows the same pattern.
```
RyanZim: I'd say just remove this comment; by now, this is going to barely resemble the original code.

lib/move-sync/__tests__/move-sync.test.js
manidlou: Very true :smile:

lib/move-sync/__tests__/move-sync.test.js
manidlou: IDK! @jprichardson?

lib/move-sync/__tests__/move-sync.test.js
manidlou: Because **prevent copying into itself** is not implemented yet for `copySync`. But as I am thinking now, that would work only for the cases if we move across devices. But, when src and dest are on the same device, we still should be able to prevent **moving into itself**. Is that right?

lib/move-sync/index.js
```diff
@@ -0,0 +1,116 @@
...
+const path = require('path')
+const copySync = require('../copy-sync').copySync
+const removeSync = require('../remove').removeSync
+const mkdirpSync = require('../mkdirs').mkdirsSync
+
+function moveSync (src, dest, options) {
+  options = options || {}
+  const overwrite = options.overwrite || options.clobber || false
+
+  if (path.resolve(src) === path.resolve(dest)) return
```
manidlou: When src and dest are the same, currently it just `return`. Which one is better, **return** or **throw an error**?

lib/move-sync/__tests__/move-sync.test.js
RyanZim: Yeah, just wanted to make sure the `.skip` wasn't included by mistake.

manidlou:
Added **prevent moving into itself** and its tests for `moveSync()`. Please let me know if there is any issues :smile:

RyanZim:
@manidlou Do you want this or https://github.com/jprichardson/node-fs-extra/pull/374 merged first? (Will cause merge conflicts either way.)

jprichardson:
Thanks @manidlou!

manidlou:
> @manidlou Do you want this or #374 merged first? (Will cause merge conflicts either way.)

Sorry for late reply as I see this already merged.

> Thanks @manidlou!

You're very welcome! :smile:

manidlou:
I will add `moveSync` docs as well.

Issue #348, opened 2017-01-26
Clarify docs for move's options 


valorloff:
```
fs.move(src,dest,[true],function(error) {
            if (error) throw error;
            console.log('Moved to '+dest);
          });
```
Error: EEXIST: file already exists   .....

RyanZim:
Docs should be clarified, correct usage of the overwrite option is: 
```js
fs.move(src,dest,{overwrite:true},function(error) {
  if (error) throw error;
  console.log('Moved to '+dest);
});
```
`options` must be an object.

Example: https://runkit.com/57ebc041d4a27f140028fd78/588a192238228e0015883bef.

valorloff:
nice!
thanks!

valorloff:
```
fs.move(oldFile, newFile,{owerwrite:true},(err)=>{
             if (err) throw err;
           });
```
```
Error: EEXIST: file already exists, link '/home/valaoffice/static/html/bla-bla.html' -> '/home/valaoffice/static/html/bla-bla-ma.html'
W20170127-11:18:11.649(2)? (STDERR)     at Error (native)
```
why?

RyanZim:
@valorloff  `link` is part of the move operation. What is the value of `oldFile` & `newFile` when this error occurs?

valorloff:
```
 if (err) throw err;   // 269
Error: EEXIST: file already exists, link '/home/valaoffice/static/html/k-t/s-p-200-m.html' -> '/home/valaoffice/static/html/d-m/s-p-200-m.html'
at Error (native)
```
moreover, moving file don't erase src file, and create new file only, this case above - exactly result of repeated fs.move

RyanZim:
What is the value of `oldFile` & `newFile` when this error occurs? Without this info, we cannot proceed to debug this.

valorloff:
console.log(src, dest);
src:  /home/valaoffice/static/html/aksessuary-dlya-mobilnyh-telefonov/Data-kabeli-adaptery/skotch-prozrachnyy-200-m.html 
dest:  /home/valaoffice/static/html/dlya-gazovoy-plity/skotch-prozrachnyy-200-m.html

RyanZim:
@valorloff I cannot reproduce this error. Please create a repository that demonstrates this error and post it to GitHub. Open a new issue with a link to the repo. Thanks!

valorloff:
I rewrite to:
```
if (!fs.existsSync(dest)) { // dest file not exist 
[...]

PR #384, opened 2017-03-09
Refactor docs to include parameters type

manidlou:
Because of the issues like #348, also to provide a little more clarification regarding parameters type. I'd like to know what you think :smile:


jprichardson:
Thanks!

Issue #391, opened 2017-03-16
Regarding Buffer


RyanZim:
A little background:

- `new Buffer(size)` is deprecated: https://nodejs.org/api/buffer.html#buffer_new_buffer_size
- `Buffer.alloc()` or `Buffer.allocUnsafe()` should be used instead, but it's not supported on Node.js v4.0.0-v4.5.0.
- https://github.com/jprichardson/node-fs-extra/pull/380 switched us to `Buffer.alloc()`. This was released in fs-extra v2.1.0
- People complained about the lack of Node v4.3.0 support: https://github.com/jprichardson/node-fs-extra/issues/390
- I reverted #380 in e02c07bc26edfca0b9602a9ac142a344ed020aad, and released fs-extra v2.1.1 as a hotfix for this.

Now, what are we going to do moving forward? Using `new Buffer()` emits a deprecation warning in Node v7.0.0-v7.2.1.

I'm wondering if we shouldn't consider some sort of fallback solution.

---

Also, this brings up the security question that prompted Node to change the Buffer API in the first place: `new Buffer` returns an uninitialized buffer. Currently, we are not zero-filling it. I can't see how this could be exploited in the current implementation, but it would be good to have this reviewed.

If we do _not_ need the buffer to be initialized, we should use `Buffer.allocUnsafe`, not `Buffer.alloc`, in later versions.

Atten: @jprichardson @manidlou @dr-dimitru

jprichardson:
To get through this mess, we should create an internal function `bufferAlloc` that checks for `Buffer.alloc` if available, and falls back to `new Buffer()` if not. Thoughts?

RyanZim:
> To get through this mess, we should create an internal function bufferAlloc that checks for Buffer.alloc if available, and falls back to new Buffer() if not.

I agree, except should we be using `alloc` or `allocUnsafe`?

jprichardson:
To err on the side of caution, my preference is `alloc` to prevent any potential security issues in the future. The only reason we'd need to use `allocUnsafe` is for performance, but it's my understanding the performance gains of `allocUnsafe` (not initializing mem to 0) are miniscule. It feels a little cargo-cultish, but I'm thinking `alloc`. This isn't a strong preference though and I could certainly be missing something in my analysis.

RyanZim:
@jprichardson I'm fine with `alloc`, but we should use `new Buffer().fill(0)` as a fallback then.

dr-dimitru:
[...]

PR #394, opened 2017-03-16
Use Buffer.allocUnsafe() in enviroments that support it

RyanZim:
Resolves https://github.com/jprichardson/node-fs-extra/issues/391

### DIFF
--- a/lib/copy-sync/copy-file-sync.js
+++ b/lib/copy-sync/copy-file-sync.js
@@ -3,7 +3,7 @@
 const fs = require('graceful-fs')
 
 const BUF_LENGTH = 64 * 1024
-const _buff = new Buffer(BUF_LENGTH)
+const _buff = require('../util/buffer')(BUF_LENGTH)
 
 function copyFileSync (srcFile, destFile, options) {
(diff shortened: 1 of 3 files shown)


RyanZim:
Should I add a small `instanceof` test case for the buffer util?

jprichardson:
> Should I add a small instanceof test case for the buffer util?

What do you mean?

jprichardson:
Oh like `instanceof Buffer` inside of the Buffer util function? Use `Buffer.isBuffer()`

manidlou:
> Should I add a small instanceof test case for the buffer util?

I am still not sure what you mean. You mean test if the returning result of the buffer util is an `instanceof Buffer`?

RyanZim:
I was wondering if I should add a testcase to `lib/util/__test__` that tests the buffer util and ensures that it returns a buffer.

manidlou:
> I was wondering if I should add a testcase to lib/util/__test__ that tests the buffer util and ensures that it returns a buffer.

Oh alright. Basically it is a good idea to have test for each feature that we add or change. However, I am not sure if this one is necessary.

PR #398, opened 2017-03-22
Don't override move test timeout

RyanZim:
This issue with graceful-fs has been fixed, so no need for this.


manidlou:
@RyanZim `moveSync()` has that timeout too: https://github.com/jprichardson/node-fs-extra/blob/master/lib/move-sync/__tests__/move-sync.test.js#L117. Would you please take care of that too? or let me know if you want me to do it :smile:

RyanZim:
Oops, yes, I'll fix that.

RyanZim:
Merging this since JP is extra busy and this change only affects tests.

Issue #401, opened 2017-04-08
Remove fs.spaces


RyanZim:
Perhaps I'm doing something wrong, but I can't get `fs.spaces` working.

Test case: https://runkit.com/57ebc041d4a27f140028fd78/58e92c306fa07e0013ccb3e6

@jprichardson ?

jprichardson:
Maybe we should remove `fs.spaces`?

RyanZim:
Suits me; it isn't documented anyhow; users who need it can use `jsonfile`. If people really complain, we can add it back.

RyanZim:
PR done: https://github.com/jprichardson/node-fs-extra/pull/402

RyanZim:
Fixed in #403; should be released in v3.0.0.

PR #402, opened 2017-04-08
BREAKING: Remove support for setting spaces for JSON writing

RyanZim:
Resolves https://github.com/jprichardson/node-fs-extra/issues/401

Note that this PR is against the new `develop` branch, which will be used for v3.0.0


lib/json/jsonfile.js
```diff
@@ -2,15 +2,21 @@
 
 const jsonFile = require('jsonfile')
 
+function write () {
+  jsonFile.writeFile.apply({ spaces: 2 }, arguments)
```
jprichardson: Seems to me it should follow the `jsonfile` default of `null` spaces. Don't you think?

RyanZim: It would definitely be simpler to do it that way; if no one's opposed, I'll change this.

RyanZim:
@jprichardson Fixed.

PR #403, opened 2017-04-10
Promise support

RyanZim:
This uses https://github.com/RyanZim/universalify, a library I wrote for this purpose.

This is a semver-major change.

~Needs docs~ (**Done!**), and we should add more tests in the future.

:tada: Promises are here! :tada:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Node.js: fs-extra
 =================
 
-`fs-extra` adds file system methods that aren't included in the native `fs` module. It is a drop in replacement for `fs`.
+`fs-extra` adds file system methods that aren't included in the native `fs` module and adds promise support to the `fs` methods. It should be a drop in replacement for `fs`.
 
 [![npm Package](https://img.shields.io/npm/v/fs-extra.svg?style=flat-square)](https://www.npmjs.org/package/fs-extra)
 [![build status](https://api.travis-ci.org/jprichardson/node-fs-extra.svg)](http://travis-ci.org/jprichardson/node-fs-extra)
@@ -30,7 +30,7 @@ Installation
 Usage
 -----
 
-`fs-extra` is a drop in replacement for native `fs`. All methods in `fs` are unmodified and attached to `fs-extra`.
+`fs-extra` is a drop in replacement for native `fs`. All methods in `fs` are attached to `fs-extra`. All `fs` methods return promises if the callback isn't passed.
 
 You don't ever need to include the original `fs` module again:
 
@@ -60,7 +60,7 @@ const fse = require('fs-extra')
 
 Sync vs Async
 -------------
-Most methods are async by default (they take a callback with an `Error` as first argument).
+Most methods are async by default. All async methods will return a promise if the callback isn't passed.
 
 Sync methods on the other hand will throw if an error occurs.
 
@@ -69,14 +69,23 @@ Example:
 ```js
 const fs = require('fs-extra')
 
+// Async with promises:
+fs.copy('/tmp/myfile', '/tmp/mynewfile')
+  .then(() => console.log('success!'))
+  .catch(err => {
+    // Handle error
+  })
+
+// Async with callbacks:
 fs.copy('/tmp/myfile', '/tmp/mynewfile', err => {
   if (err) return console.error(err)
-  console.log("success!")
-});
+  console.log('success!')
+})
 
+// Sync:
 try {
   fs.copySync('/tmp/myfile', '/tmp/mynewfile')
-  console.log("success!")
+  console.log('success!')
 } catch (err) {
   console.error(err)
 }
@@ -119,7 +128,7 @@ Methods
 - [writeJsonSync](docs/writeJson-sync.md)
 
 
-**NOTE:** You can still use the native Node.js methods. They are copied over to `fs-extra`.
+**NOTE:** You can still use the native Node.js methods. They are promisified and copied over to `fs-extra`.
 
 ### What happened to `walk()` and `walkSync()`?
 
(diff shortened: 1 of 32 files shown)


RyanZim:
Need to fix my hacky tests since they fail on Node 4 & 5.

RyanZim:
Added basic README docs; should eventually migrate `docs/` in a seperate PR.

README.md
```diff
@@ -69,11 +69,20 @@ Example:
 ```js
 
+// Async with promises:
+fs.copy('/tmp/myfile', '/tmp/mynewfile')
+  .then(() => console.log("success!"))
```
manidlou: Probably not that important, but for the purpose of consistency in using `'`, wouldn't it be better `console.log("success!")` replaced with `console.log('success!')` since we use `'` pretty much everywhere else? It actually has been there since the beginning (I guess). I just saw it now and thought that's be nice that we fix it here :smile:

README.md
```diff
@@ -69,11 +69,20 @@ Example:
...
 fs.copy('/tmp/myfile', '/tmp/mynewfile', err => {
   if (err) return console.error(err)
   console.log("success!")
-});
+})
 
+// Sync:
 try {
   fs.copySync('/tmp/myfile', '/tmp/mynewfile')
   console.log("success!")
```
manidlou: Same here: `console.log("success!")` -> `console.log('success!')`?

README.md
```diff
@@ -69,11 +69,20 @@ Example:
...
+fs.copy('/tmp/myfile', '/tmp/mynewfile')
+  .then(() => console.log("success!"))
+  .catch(err => {
+    // Handle error
+  })
+
+// Async with callbacks:
 fs.copy('/tmp/myfile', '/tmp/mynewfile', err => {
   if (err) return console.error(err)
   console.log("success!")
```
manidlou: Also here, `console.log("success!")` -> `console.log('success!')`?

manidlou:
Other than that minor, LGTM. Nice work @RyanZim :clap:

README.md
RyanZim: Yeah, we should fix that; thanks for the catch. I also removed an old semicolon in these docs. :wink: 

Someone ought to make a tool to run eslint on code blocks in markdown.

RyanZim:
@manidlou Fixed.

---

Should also add: I'd appreciate review/improvements to https://github.com/RyanZim/universalify as well.

jprichardson:
Nice work @RyanZim! 

1) Will you verify that the test cases for universalify include any additional test cases from https://github.com/sindresorhus/pify/blob/master/test.js and https://github.com/paulmillr/micro-promisify/blob/master/test.js - i.e. I'm just trying to ensure unverisalify doesn't have any edge case bugs that these battle-hardened libraries have already solved.

2) Maybe I missed it, but will you add some tests demonstrating that a method imported from `fs` returns a promise?

Thanks!

RyanZim:
@jprichardson 

1. All the cases covered by pify, are already tested; except pify works on entire modules, which is something universalify will never do.
    
    micro-promisify has a few cases I should add. Thanks for the suggestions!
1. Already done in `lib/fs/__tests__/mz.test.js`

manidlou:
> Someone ought to make a tool to run eslint on code blocks in markdown.

I believe that's a great idea!

Actually, that is specifically mentioned in [standardjs](https://github.com/feross/standard) docs:

> [Can I check code inside of Markdown or HTML files?](https://github.com/feross/standard#can-i-check-code-inside-of-markdown-or-html-files)

As it is mentioned there, they introduce two modules for that:

- [standard-markdown](https://github.com/zeke/standard-markdown)
- [eslint-plugin-markdown](https://github.com/eslint/eslint-plugin-markdown)

I tried `standard-markdown` (on our old `README.md`), and this was the result:

```
Linting file 1 of 1

   README.md
         74:15:    Strings must use singlequote.
         75:3:     Extra semicolon.
         79:15:    Strings must use singlequote.
```

Then, we can add a new `scripts` something like `lintmd` to `package.json` to automate this process.

So, @jprichardson, @RyanZim what do you think?

RyanZim:
@manidlou Yeah, new issue or PR for that.

RyanZim:
@jprichardson micro-promisify sets the `name` and `length` properties on the returned function. I'm fine with setting `name`, but setting `length` isn't too straightforward, since we are potentially adding or removing the final callback parameter. I'd just skip setting `length` IMO. Are you OK with that?

RyanZim:
Updated `docs/` to reflect this change.

louisgv:
LGTM 👍

RyanZim:
@jprichardson Updated PR to use universalify v0.1.0, which sets `name` for the returned function.

Anything else before the merge?

jdalton:
Hurray! I was using `pify` to wrap fs-extra methods 😸

rijnhard:
Hi

This is pretty cool because now i dont need `fs-extra-promise` but its also kind of annoying that its opinionated. For instance I use `Q` for pretty much everything and now I have to wrap any `fs` returns like this `Q(fs.close())`.
Ideally using a library like `any-promise` makes it easy for the user to pick his preferred implementation.

RyanZim:
@rijnhard I'm of the opinion that fs-extra should be as light as possible. Adding any-promise increases weight. I'm firmly opposed to this ATM.

rijnhard:
@RyanZim theres another alternative I've seen, where theres some variable or method that accepts an A+ promise implementation, then they use that for all the promise creation. 
like

```javascript
import Q from 'q';
import fs from 'fs-extra';
fs.setPromise(Q.Promise);
```
and in the code you use the standard es6 promise as you would.

RyanZim:
Issue filed: https://github.com/jprichardson/node-fs-extra/issues/425

Issue #145, opened 2015-07-02
Add `fs.exists` and `fs.existsSync` if it doesn't exist.


jdalton:
Since `fs.exists` and `fs.existsSync` [are deprecated](https://github.com/nodejs/io.js/issues/103) it would be good to check if they exist and add them if they don't. You could reference https://github.com/sindresorhus/path-exists.

jprichardson:
I've considered this, but haven't acted because it's my understanding that despite the fact that they're being deprecated, they're still currently included in io.js and the deprecation message hasn't been enabled yet (or was disabled - https://github.com/nodejs/io.js/issues/257). So I didn't feel any sense of urgency. Do you know if there's any discussion on actually removing the methods? Regardless, I suppose it wouldn't hurt to be ready.

jdalton:
>  Do you know if there's any discussion on actually removing the methods?

I do not.

> Regardless, I suppose it wouldn't hurt to be ready.

Yap. I figure if it were added then when they do pull the trigger fs-extra would be ready.

goloroden:
There was a reason why they were deprecated: It misleads developers to rely on them to ensure that a file exists, but it directly leads to a race condition.

Hence, it's basically a good thing that they have been deprecated. I'm not sure if `fs-extra` should re-enable them, with respect to the fact that there is an actual reason why they have been deprecated.

Wouldn't it be better to help spread the word on how to *correctly* deal with situations where you would use `fs.exists`? Maybe I'm missing something, so feel free to correct me if you think that I'm wrong… I was just thinking out loud ;-).

jdalton:
I think it's handy and worst case document the possible race condition. I suspect it doesn't apply to most devs using it so a doc note would do.

jprichardson:
> Wouldn't it be better to help spread the word on how to correctly deal with situations where you would use fs.exists? Maybe I'm missing something, so feel free to correct me if you think that I'm wrong… I was just thinking out loud ;-).

I think the core Node devs are wrong on this one. No doubt that they're right that it *could* lead to a race condition if you check if a file exists before opening it. But sometimes you just want to know if a file exists or not. Not because you want to read/write it, but because the mere existence may indicate the completion of some operation. 

That being said, they're very useful methods and `fs-extra` will add them. However, I won't hide the original `exists/existsSync` methods, I'll need to come up with new names. I'll probably change the signature too from `(doesExist)` to `(error, doesExist)` so that this entire library can easily be promisified.

[...]

PR #406, opened 2017-04-24
Add pathExists()

RyanZim:
This is my approach at resolving https://github.com/jprichardson/node-fs-extra/issues/145.

`fs.exists` is broken. We're a drop-in replacement for `fs`, so we can't fix it.

We can make a new method that acts like `fs.exists` should have behaved all along. I've named it `pathExists()`.

I also aliased `fs.existsSync` to `fs.pathExistsSync` for consistency.

## Example Usage
```js

const file = '/tmp/this/path/does/not/exist/file.txt'
// Promise usage:
fs.pathExists(file)
  .then(exists => console.log(exists)) // => false
// Callback usage:
fs.pathExists(file, (err, exists) => {
  console.log(err) // => null
  console.log(exists) // => false
})
```


manidlou:
LGTM! :+1:

jprichardson:
I like it, nice work @RyanZim!

jaredallard:
I'm interested in why `fs-extra` doesn't support using the not-broken `fs.exists` when accessing it via a promise. Seems like you could make it drop-in compatible via the promise version and keep using the broken version when NOT accessing it via a promise. Maybe I'm missing something though

RyanZim:
@jaredallard We do promisify `fs.exists`, and still offer the broken `fs.exists` to callback users. However, we encourage the use of `fs.pathExists`, since it doesn't use deprecated Node APIs, and the callback version behaves correctly.

jaredallard:
@RyanZim Yes, I know `fs.exists` is promisified, but I'm wondering why not have it point to/use the solution that `pathExists` implements, since you can literally rewrite `fs.exists -> fs.pathExists` right now. Since how `fs.exists` behaves (in it's broken way) is somewhat how it's consumed as a promise (`catch` is used as the `err` instead). 

**TL;DR**
The normal callback way _should_ point to the broken one for compat, but the `fs.exists` _promise_ version could be the same as the current `fs.pathExists` AFAIK.

PR #411, opened 2017-04-29
Add standard-markdown to lint js code in markdown files

manidlou:
Added [standard-markdown](https://github.com/zeke/standard-markdown) to lint js code blocks inside our markdown files. Since `standard-markdown` uses `standard`, we have a consistent linter for both js and markdown files.

`standard-markdown` by default comes with a few eslint rules disabled. Just added one more rule [handle-callback-err](http://eslint.org/docs/rules/handle-callback-err) to the list of `disabledRules` in `standard-markdown` source code. I guess for us, it makes sense to disable that rule on markdown files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,9 +72,7 @@ const fs = require('fs-extra')
 // Async with promises:
 fs.copy('/tmp/myfile', '/tmp/mynewfile')
   .then(() => console.log('success!'))
-  .catch(err => {
-    // Handle error
-  })
+  .catch(err => console.error(err))
 
 // Async with callbacks:
 fs.copy('/tmp/myfile', '/tmp/mynewfile', err => {
--- a/docs/copy.md
+++ b/docs/copy.md
@@ -35,7 +35,7 @@ fs.copy('/tmp/myfile', '/tmp/mynewfile')
   console.log('success!')
 })
 .catch(err => {
-  // handle error
+  console.error(err)
 })
 ```
 
--- a/docs/emptyDir.md
+++ b/docs/emptyDir.md
@@ -25,6 +25,6 @@ fs.emptyDir('/tmp/some/dir')
   console.log('success!')
 })
 .catch(err => {
-  // handle error
+  console.error(err)
 })
 ```
--- a/docs/ensureDir.md
+++ b/docs/ensureDir.md
@@ -24,6 +24,6 @@ fs.ensureDir(dir)
   console.log('success!')
 })
 .catch(err => {
-  // handle error
+  console.error(err)
 })
 ```
--- a/docs/ensureFile.md
+++ b/docs/ensureFile.md
@@ -24,6 +24,6 @@ fs.ensureFile(file)
   console.log('success!')
 })
 .catch(err => {
-  // handle error
+  console.error(err)
 })
 ```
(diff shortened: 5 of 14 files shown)


manidlou:
Aah! I have to work on it a little more. Need to find a way to programmatically add `disabledRules` to `standard-markdown`. Maybe I should fork it and apply changes based on our needs!

manidlou:
So Added `tools` directory to contain `standard-markdown` fork. Updated `.gitignore` and `.npmignore` accordingly. Please let me know if I am doing anything incorrectly here!

manidlou:
Alright now it's fixed! :grin:

RyanZim:
Hey @manidlou; thanks for putting in the effort here.

I don't like the idea of maintaining a fork here though. I think perhaps we should propose that standard-markdown disables this rule globally, since it does make sense in some cases.

How many cases does this rule fail? If it's not too much work, I wouldn't mind changing a few examples to work around this.

manidlou:
Thanks @RyanZim. I like your approach toward this since I essentially propose this as an idea

> I don't like the idea of maintaining a fork here though.

Yeah, I am with on that :+1: 

> I think perhaps we should propose that standard-markdown disables this rule globally, since it does make sense in some cases.

That or propose the idea of make `disabledRules` configurable by the user; something like a property in `package.json` similar to `standard`.

> How many cases does this rule fail? If it's not too much work, I wouldn't mind changing a few examples to work around this.

Currently, when I run it without disabling `handle-callback-err`, 16 of them get this error. Here is the output copied from my terminal:

```
Linting file 30 of 30


   docs/copy.md
         37:8:     Expected error to be handled.

   docs/emptyDir.md
         27:8:     Expected error to be handled.

   docs/ensureDir.md
         26:8:     Expected error to be handled.

   docs/ensureFile.md
         26:8:     Expected error to be handled.

   docs/ensureLink.md
         26:8:     Expected error to be handled.

   docs/ensureSymlink.md
         27:8:     Expected error to be handled.

   docs/move.md
         26:8:     Expected error to be handled.

   docs/outputFile.md
         19:29:    Expected error to be handled.
         30:8:     Expected error to be handled.

   docs/outputJson.md
         22:21:    Expected error to be handled.
         33:8:     Expected error to be handled.

   docs/readJson.md
         28:8:     Expected error to be handled.
         55:8:     Expected error to be handled.

   docs/remove.md
         31:8:     Expected error to be handled.

   docs/writeJson.md
         29:8:     Expected error to be handled.

   README.md
         75:10:    Expected error to be handled.
```

manidlou:
For now, I will fix those cases.

RyanZim:
@manidlou Can you rebase this into one commit? We don't want the fork in the git history.

manidlou:
> @manidlou Can you rebase this into one commit? We don't want the fork in the git history. :+1: 

> One thing I just noticed; your Travis config is running more containers than needed. Why don't you just add standard-markdown to the lint script & revert your .travis.yml changes? :+1: 

Sure thing. Just don't have access to computer now. I'll do them when I go home tonight.

manidlou:
Rebase is done. Please let me know if I missed anything :stuck_out_tongue_winking_eye:

jprichardson:
Looks good @manidlou, thanks!

PR #412, opened 2017-05-01
Provide optional content when ensuring file.

hovosanoyan:


RyanZim:
How is this different from `outputFile`? I don't see why you can't just use that.

Closing for now, will reopen if there is a clear case presented for this.

hovosanoyan:
The reason I suggested this, is that I came across with a case when I need to append data to the file only if it don't exist.

RyanZim:
Ah, OK. I understand now. This seems like a very small edge case to support, but I'd like @jprichardson & @manidlou's thoughts here.

If we were to go ahead adding this, we'd need the async version changed as well, along with docs and tests.

jprichardson:
NACK. It's super confusing. If you read:

```js
fs.ensureFileSync(someFile, 'this is some content')
```

It reads as if it will always "ensure" the content is `this is some content`. But that case is only valid if the file doesn't exist.

RyanZim:
I agree with @jprichardson.

It's a lot clearer to do:
```js
if (!fs.pathExists(someFile)) fs.outputFile(someFile, 'content')
```
Closing since the majority of the team is against this.

manidlou:
Sorry for being late. Definitely agreed on the decision.

PR #415, opened 2017-05-02
move() should error when src & dest are the same and src doesn't exist

### LINKED ISSUE #414: fs.move doesn't return error on moving same nonexistent folder/file
#### Issue:
When i `fs.move()` using two paths that are the same and both are invalid(nonexistent) paths, the function does not return an error, which it should since you are trying to move a nonexistent path.

#### Setup:
This is being ran inside of an electron render, which means its being ran inside of a chrome wrapper. This issue persists through fs-extra 3.0

#### Example:
`fs.move("nonexistent/path","nonexistent/path", (err) => { console.log(err)})`
`// err returns null`

#### Usage:
I use this in my personal repo:
https://github.com/zosman1/The_Archiver/blob/c65ed9dfb7ae0855b1fcb8026204c8d6876a96af/app/index.js#L25

RyanZim:
Fixes #414; bug introduced in #378

@jprichardson What's your opinion on semver for this patch?

### DIFF
--- a/lib/move-sync/__tests__/move-sync.test.js
+++ b/lib/move-sync/__tests__/move-sync.test.js
@@ -61,6 +61,13 @@ describe('moveSync()', () => {
     assert.ok(contents.match(expected), `${contents} match ${expected}`)
   })
 
+  it('should error if src and dest are the same and src does not exist', () => {
+    const src = `${TEST_DIR}/non-existent`
+    const dest = src
+
+    assert.throws(() => fse.moveSync(src, dest))
+  })
+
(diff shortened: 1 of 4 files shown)


manidlou:
Thanks @RyanZim :+1:

RyanZim:
@jprichardson ping?

RyanZim:
@jprichardson Want this released? What semver?

jprichardson:
> @jprichardson Want this released? What semver?

yeah, `3.0.1`

PR #419, opened 2017-05-06
Improve filter option in copy()

manidlou:
I believe `filter` option is a very useful option. However, it currently doesn't behave consistently for all cases. One example is #350. Therefore, it requires extra efforts from users to achieve what they want. Sometimes it's so difficult to get what you want simply because of a complex directory structure and/or maybe you even don't have any knowledge of the src directory structure. I believe there should be a way to provide a `filter` option that behaves consistently and intuitively for all cases and with minimal efforts from users.

In general, one use case that `filter` option comes very handy is to copy certain files (and most of the time we want it recursively). To clarify the issue, consider these examples. The goal is to copy only `.js` files. With existing behavior, none of them work as expected.

```
src
|__foo
|  |__file1.js
|
|__bar
|  |__file2.js
|
|__baz
   |__file3.md
```

**Problem: nothing copied** (since src fails on `filter`)
```js
const path = require('path')

const fn = s => path.extname(s) === '.js'

fs.copy('src', 'dest', {filter: fn}, err => {
  console.log('done.')
})
```

**Problem: empty dest directory**
```js
const path = require('path')

const src = path.resolve('src')
const fn = s => path.extname(s) === '.js' || s === src

fs.copy(src, 'dest', {filter: fn}, err => {
  console.log('done.')
})
```
**Problem: empty `baz` directory**
```js
const path = require('path')

const fn = s => path.extname(s) === '.js' || fs.lstatSync(s).isDirectory()

fs.copy('src', 'dest', {filter: fn}, err => {
  console.log('done.')
})
```
Here is my proposal to provide a consistent `filter` option. (_side note: it is how I handled this issue in [klaw-sync](https://github.com/manidlou/node-klaw-sync#usage) since that suffered from this issue as well, however here is more involved though._)

I agree it may make the code a little complicated, but it gives us more robust functionality instead. Essentially, the behavior is the same when `filter` passes successfully. The issue arises when `filter` fails. So, as a default behavior, for directories that fail on `filter`, don't copy them but read (recurse) them to see if their contents pass `filter`. Although I am not a fan of functions with several options! but we also provide an additional `boolean` option something like `noRecurseOnFailedFilter` if users want to change this default recursive behavior based on their needs.

With these changes, users can use `filter` option as simple as this and expect to get the correct results:
```js
const path = require('path')

const fn = s => path.extname(s) === '.js'

fs.copy('src', 'dest', {filter: fn}, err => {
  console.log('done.')
})
```

Please feel free to point out any issues with this argument.

If you think it makes sense and it is somehow acceptable, I will apply these changes to `copySync()` and #374 as well.

### DIFF
--- a/lib/copy/ncp.js
+++ b/lib/copy/ncp.js
@@ -3,6 +3,8 @@
 var fs = require('graceful-fs')
 var path = require('path')
 var utimes = require('../util/utimes')
+var mkdirp = require('../mkdirs').mkdirs
+var pathExists = require('../path-exists').pathExists
 
 function ncp (source, dest, options, callback) {
   if (!callback) {
@@ -15,6 +17,7 @@ function ncp (source, dest, options, callback) {
   var targetPath = path.resolve(basePath, dest)
 
   var filter = options.filter
+  var noRecurseOnFailedFilter = options.noRecurseOnFailedFilter
   var transform = options.transform
   var overwrite = options.overwrite
   // If overwrite is undefined, use clobber, otherwise default to true:
@@ -34,18 +37,6 @@ function ncp (source, dest, options, callback) {
 
   function startCopy (source) {
     started++
-    if (filter) {
-      if (filter instanceof RegExp) {
-        console.warn('Warning: fs-extra: Passing a RegExp filter is deprecated, use a function')
-        if (!filter.test(source)) {
-          return doneOne(true)
-        }
-      } else if (typeof filter === 'function') {
-        if (!filter(source, dest)) {
-          return doneOne(true)
-        }
-      }
-    }
     return getStats(source)
   }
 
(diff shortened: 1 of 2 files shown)


lib/copy/ncp.js
```diff
@@ -65,13 +56,80 @@ function ncp (source, dest, options, callback) {
...
+            })).then(() => callback()).catch(callback)
+          })
+        }
+
+        function mkParentDir () {
+          var destParent = path.dirname(dest)
+          pathExists(destParent, (err, dirExists) => {
+            if (err) return callback(err)
+            if (dirExists) return onDir(item)
+            mkdirp(destParent, err => {
```
RyanZim: No need for this, `mkdirp` won't error if the directory already exists. Just call `mkdirp` without checking if it exists.

RyanZim:
I sorta like this, except I'm just wondering if `noRecurseOnFailedFilter` is needed. Or can we just change the behavior and release as a major version?

manidlou:
> I'm just wondering if `noRecurseOnFailedFilter` is needed.

That'd be great if we could do this without an extra option. Essentially, you still get correct results without using this option. This option can be useful when you care about performance.

This may not necessarily true, but I guess this option can be useful when you use `filter` to exclude directories rather than include them.

For example, if you use `filter` to exclude `node_modules`:
```js

const fn = s => s.indexOf('node_modules') < 0

fs.copy('src', 'dest', {filter: fn}, err => {
  console.log('done.')
})
```
This case (and similar cases), although you still get correct results without using `{noRecurseOnFailedFilter: true}`, you still have a wasteful operation of traversing unwanted directories. Therefore, in this case, you want to use `{noRecurseOnFailedFilter: true}` since you know you don't want anything from `node_modules` (no files, no directories).

That's one thing that I can think of regarding that option. Other than that, hey `less options => less confusion` :wink: !

manidlou:
> We could release the rewrite of ncp along with this. :+1: 

I work on applying these changes to #374.

manidlou:
Applied these changes to #374. Please see https://github.com/jprichardson/node-fs-extra/pull/374/commits/6271e98628b52d88b252913d385f516ead6db5e6.

PR #422, opened 2017-05-07
Add filter option to remove() and removeSync()

manidlou:
Added `filter` option to `remove()` and `removeSync()`. I am not sure if you like it. But, I figured it's worth of proposing.

This `filter` option accepts a function that returns `true` to include (remove) and `false` to exclude (not remove). The same as `copy()`. It operates recursively as well. 

I think `filter` option can be useful for `remove` as well. One useful example is when you want to remove only certain files and/or directories. For instance, with this option, users can run something as simple as this to remove only `.js` files:

```js
const path = require('path')

const fn = p => path.extname(p) === '.js'

fs.remove('/some/dir', {filter: fn}, err => {
  console.log('done.')
})
```

I'd like to know your thoughts on this please! :grin:

### DIFF
--- a/lib/remove/__tests__/remove-filter.test.js
+++ b/lib/remove/__tests__/remove-filter.test.js
@@ -1,0 +1,118 @@
+'use strict'
+
+const assert = require('assert')
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+const fse = require(process.cwd())
+
+/* global afterEach, beforeEach, describe, it */
+
+let TEST_DIR
+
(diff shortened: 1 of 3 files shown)


RyanZim:
I hesitate to add this, since the remove code is a fork of `rimraf`, and I'm not sure if we should diverge too far from upstream.

Also, I'm of the opinion that you shouldn't add a feature unless you or a user _needs_ it.

manidlou:
Fair enough! Will be closed then.

PR #423, opened 2017-05-08
Improve filter option in copy()

manidlou:
Ref: Started #419.

Renamed branch to `filter-copy` to be more specific.

Also removed `noRecurseOnFailedFilter` for the sake of simplicity. If you think still need it, I will add it again.

Just a quick summary:

I believe `filter` option is a very useful option. However, it currently doesn't behave consistently for all cases. One example is #350. Therefore, it requires extra efforts from users to achieve what they want. Sometimes it's so difficult to get what you want simply because of a complex directory structure and/or maybe you even don't have any knowledge of the src directory structure. I believe there should be a way to provide a filter option that behaves consistently and intuitively for all cases and with minimal efforts from users.

In general, one use case that `filter` option comes very handy is to copy certain files and/or directories, and most of the time we want it recursively.

With these changes, users can use `filter` option as simple as this and expect to get the correct results:
```js
const path = require('path')

const fn = s => path.extname(s) === '.js'

fs.copy('src', 'dest', {filter: fn}, err => {
  console.log('done.')
})
```

### DIFF
--- a/lib/copy/__tests__/copy.test.js
+++ b/lib/copy/__tests__/copy.test.js
@@ -324,7 +324,7 @@ describe('fs-extra', () => {
         }, 1000)
       })
 
-      it('should apply filter recursively on failed dirs by default', done => {
+      it('should apply filter recursively on failed dirs', done => {
         const srcDir = path.join(TEST_DIR, 'src')
         const srcFile1 = path.join(srcDir, '1.js')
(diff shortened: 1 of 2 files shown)


manidlou:
closing this since `filter` option is going through major changes in rewrite of `copy`.

PR #424, opened 2017-05-08
Improve filter option in copySync

manidlou:
Applied the improved `filter` option that introduced in #423 to `copySync()`.

Also cleaned up `copySync` a bit including:

- removed `recursive` option. This option never been documented and had no use.
- Add `copyFileSync()` as a function and wrap everything in one file since the entire file is just 114 lines!

In addition, removed `option.overwrite = false` in `moveSync()` since one of the tests (when move directory and overwrite is true) wouldn't pass. It was weird though as it would pass before! :confused: IDK! maybe because of the closure! After all, I guess it's better to remove that line though. Please correct me if I am wrong :grimacing:


manidlou:
closing this since `filter` option is going through major changes in rewrite of `copy` that will affect `copySync` as well because of consistency.

Issue #430, opened 2017-05-23
flush file system on copy


Inve1951:
i have to ensure that the files copied using `fse.copy()` are actually written to the physical drive.
i'm proposing a flag for it in the function's options object.

this could be achieved by using `fs.fsync()` or possibly even by simply listening for the `'finish'` event instead of the `'close'` event.

second way sure is really easy to implement. however i'm unsure what ["has been flushed to the underlying system"](https://nodejs.org/api/stream.html#stream_event_finish) actually means.

if it flushes to windows' write cache then it's unsufficient.

RyanZim:
`copy()` used to use `finish`, see https://github.com/jprichardson/node-fs-extra/issues/326 for discussion.

Inve1951:
too bad.

just experimented a bit and found out that `fs.createWriteStream()`'s returned `WriteSteam` object comes with the property `fd` which can be successfully used with `fsync()`.

it would have to be called before the `'end'` event since the file descriptor is already closed when it's emitted.

node.js docs reffer to [this](http://man7.org/linux/man-pages/man2/fsync.2.html) for `fsync(fd)`:
> fsync() transfers ("flushes") all modified in-core data of (i.e.,
       modified buffer cache pages for) the file referred to by the file
       descriptor fd to the disk device (or other permanent storage device)
       so that all changed information can be retrieved even after the
       system crashed or was rebooted.  This includes writing through or
       flushing a disk cache if present.

but
> Calling fsync() does not necessarily ensure that the entry in the
       directory containing the file has also reached disk.  For that an
       explicit fsync() on a file descriptor for the directory is also
       needed.

so after finished writing and flushing all files of a directory we'd have to do another `fsync()` on the directory.

Inve1951:
wrote #432 as a starting point.
it seems to do what it should. i got no idea how to confirm that however.

RyanZim:
@Inve1951 Have you actually verified that fs-extra as it currently exists is causing an issue here?

Inve1951:
@RyanZim I'm not sure what issue you are refering too. This is about adding a feature.
Flushing the copied content is required to ensure it's intact after e.g. a power failure. (think backup system)

Inve1951:
@RyanZim, @jprichardson

was able to confirm the PR's functionallity.

tested in a VM with WinXP SP3 x86 NTFS, both OS and disk write cache enabled.
forced a bluescreen after the callback returned. (immediately and after a few secs delay)

[...]

PR #432, opened 2017-05-26
[draft] flush (OS) write cache with fse.copy() using options.sync

Inve1951:
flushes all copied files and directories including the destination base but not any parent directories (which might be desired if destination parent didn't exist)


RyanZim:
Closing as per https://github.com/jprichardson/node-fs-extra/issues/430#issuecomment-304467128

PR #444, opened 2017-06-20
Support util.promisify on node>=8 (Fixes issue #442)

### LINKED ISSUE #442: Move to using `util.promisify` to return correct value in promise
Node.js 8 has `util.promisify`, which calls the function at the `util.promisify.custom` symbol to return a custom promise.
This is used in the Fs module to return different values from the promise. 

For example, in the `read` method, instead of an array of the callback arguments, an object will be returned like `{ bytesRead: 1234, buffer: <Buffer 0x12 0x34 0x56 0x78> }`.

As `fs-extra` intends to be fs-compliant, it should return the correct promise also according to the docs at https://nodejs.org/api/fs.html.

Note:
This may introduce some breaking changes where promises in fs-extra returned arrays previously. Should be a major version release for moving to Node.js 8.

danielgindi:


danielgindi:
Seems like the CI has some trouble running node 6, and sometimes node 4. I tried 3 times...

RyanZim:
@danielgindi I'm not sure I like this approach. I think fs-extra should return objects like `util.promisify` by default.

danielgindi:
"By default" you mean when calling `Fs.read(...).then...` as opposed to util.promisify(Fs.read)(...).then...`?

If so - then it is easily fixable, and I started out this way. But then I had backwards compatibility in mind.
So if you're going to have a major version release - it's a good idea to always return the node-v8 version of things.

We could decide to return it "by default" but only on node-v8, but it's still a breaking change, as opposed to the node-v8 change which is just an addition.

Let me know what you think.

RyanZim:
@danielgindi I mean that you should be able to call `fs.read(...).then...` and get an object in every Node version from v4-8, while still being able to call `util.promisify(fs.read)(...).then...` in Node 8.

`fs-extra` must be a drop-in replacement for the `fs` module. If you have code that works, you should be able to change `require('fs')` to `require('fs-extra')` and the code should still work. That's the basic rule here.

At the same time, we want `fs-extra` to give developers an awesome experience, far better than working with the native `fs` module. Wrapping every fs method in `util.promisify()` isn't an awesome experience IMO. That's why `fs-extra` has built-in promise support.

Also see discussion at https://github.com/ljharb/util.promisify/issues/4, this could help make this easier.

danielgindi:
There's something very unclear here. Seems like a misunderstanding:

1. `util.promisify(fs.read)(...).then.` in the PR now works as it should (returning object)
2. `fs.read(...).then...` works like in previous versions, backwards compatible. *node v0-8 versions do not support this syntax anyway.`. Should we make a breaking change and return an object too?
3. What the heck do we need `util.promisify` for? We detect that there's a `util.promisify`, and add the promisify functionality. This happens on node-v8 only, which is the only version that has a special object returned anyway.

RyanZim:
> Should we make a breaking change and return an object too?

Yes. I think is reasonable to guess that sometime down the road, Node will add promises to core and that those promises will return objects.

danielgindi:
Okay, I'll make those return a `promisify` promise by default, and rebase :-)

RyanZim:
@jprichardson and I have talked this over and with a bit more consideration, we've decided to hold back on switching to `util.promisify`, at least in the short term.

Polyfilling graceful-fs seems like a hacky solution to this, and could cause problems due to the global nature of CommonJS modules.

At this time, I think we'll keep using universalify, adding a custom implementation for `fs.read` & `fs.write`, like we're already doing for `fs.exists`. 

I'm sorry about this; I hate having to turn down your hard work. Thanks for the effort!

danielgindi:
Well, that's a shame.

PR #445, opened 2017-06-20
Removed extra Promise and extra callback

danielgindi:
These lines were creating a promise, wrapping it in a callback.
While we can just use the original callback.


danielgindi:
Irrelevant anymore, happens as part of https://github.com/jprichardson/node-fs-extra/pull/444

Issue #436, opened 2017-05-31
Return Objects from fs.read & fs.write


teledemic:
The promisified version of fs.read shows up in the typescript typings, but when run gives "fs.read is not a function". I'm using the latest versions of both:
```
		"fs-extra": "3.0.1",
		"@types/fs-extra": "3.0.2",
```
and calling as I would expect to:
```
let [bytesread, buffer] = await fs.read(fd, buf, 0, length, start);
```

RyanZim:
This is due to the way our promisification logic works (or rather, doesn't work). The promise only returns `bytesread`, not the buffer. So usage would be:

```js
let buf = Buffer.alloc(length)
let bytesread = await fs.read(fd, buf, 0, length, start)
// Use buf here
```

The reason for this is that `fs.read` calls the callback with 3 arguments, but our promisification logic only passes the second argument in the promise.

This is sub-optimal and should probably be fixed in the next major release. However, I'm thinking we should follow the lead of `util.promisify`'s behavior. From the docs:

> If this method [`fs.read()`] is invoked as its `util.promisify()`ed version, it returns a Promise for an object with `bytesRead` and `buffer` properties.

CC: @jprichardson @manidlou

teledemic:
Thanks! That works great, and having two references to the same buffer was redundant anyway. The typescript defs are still showing it returning [number, Buffer], but that's not an issue for this repo.

RyanZim:
Yeah, could submit a fix here: https://github.com/DefinitelyTyped/DefinitelyTyped

PR #449, opened 2017-06-26
BREAKING: fs.read() & fs.write() should return objects

RyanZim:
For forward-compat with `util.promisify()`

Resolves https://github.com/jprichardson/node-fs-extra/issues/436

Needs docs yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Methods
 - [writeJsonSync](docs/writeJson-sync.md)
 
 
-**NOTE:** You can still use the native Node.js methods. They are promisified and copied over to `fs-extra`.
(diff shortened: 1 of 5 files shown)


RyanZim:
Windows CI fails due to funky windows timestamps; unrelated to this change.

yamboy1:
@RyanZim But why does the node.js v4 build pass and the v6 build fail?

RyanZim:
@yamboy1 It's a 32-bit windows issue; nothing critical.

yamboy1:
@RyanZim It's reasons like this why I prefer to use Linux

yamboy1:
Windows just has too many bugs

PR #450, opened 2017-06-27
Skip copy & copySync timestamp tests on 32-bit computers

RyanZim:
This should get appveyor passing


RyanZim:
> Looks good, but for full transparency, IIRC, the arch property doesn't actually tell if the CPU processor is 32 or 64, but only if Node was compiled with 32-bit / 64-bit. I.e. you could have 32-bit Node.js running on a 64-bit operating system.

Now that you say this, I do remember talking about this when we added a warning for using the `preserveTimestamps` option. IIRC, if Node is compiled for 32-bit, timestamps still aren't accurate. Anyhow, doesn't really matter since this is just tests.

PR #451, opened 2017-06-27
specify minimum node version (4.5)

justinhelmer:
With node 4.4 or earlier, tar fails on `Buffer.alloc` which was added in [4.5](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V4.md#4.5.0)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
     "type": "git",
     "url": "https://github.com/jprichardson/node-fs-extra"
   },
+  "engines": {
+    "node": ">=4.5"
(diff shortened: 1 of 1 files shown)


RyanZim:
We will be supporting Node 4.* until Node 4 LTS is EOLed. I'm willing to look into any problems on older Node 4 versions.

RyanZim:
See https://github.com/jprichardson/node-fs-extra/issues/452#issuecomment-311415793

PR #458, opened 2017-07-10
Prevent move into itself

manidlou:
Fixed the issue with `move()` when dest is a subdirectory of src. Also, removed `shouldMkdirp` option as we discussed before.

To quickly remind the issue, with existing behavior, running something like this
```js
const src = 'src'
const dest = 'src/dest'
fs.move(src, dest, err => {
  console.log('done.')
})
```
results in just removing `src` and essentially losing all `src` contents.

I believe we should prevent this from happening. Any feedback appreciated!


jprichardson:
Stellar work @manidlou! @RyanZim please publish a patch when you're able.

PR #463, opened 2017-07-17
Fix changelog move() notes

manidlou:
Fixed `CHANGELOG.md` latest `move()` notes.


RyanZim:
@manidlou Good catch; what was I thinking? :confused:

PR #473, opened 2017-07-31
Update & Optimize CI configs

RyanZim:
- Test on Node 8
- Only run unit tests on Appveyor
- Optimize TravisCI builds to use the minimum number of containers while keeping the performance reasonable


RyanZim:
Travis takes somewhere between 1-1.5 min. to run; AppVeyor still takes ~3 min. due to no concurrent builds.

PR #474, opened 2017-08-10
wrap function added

rafasoyyo:
Function to wrap the content of a folder inside a new folder


RyanZim:
Not sure if this is an edge case worth supporting or not. If you need to perform this operation, yes, it's painful. However, I'm not sure that this operation is common enough to warrant inclusion in fs-extra. @jprichardson @manidlou ?

If we do decide to include this, the code needs cleanup and refactoring, and the method needs documenting.

jprichardson:
Thanks for the contribution @rafasoyyo - but this is too much of an edge case for our tastes, so we'll pass.

Issue #478, opened 2017-08-17
[Windows] fs.copy error with {preserveTimestamps}


laggingreflex:
`fs.copy(src, dest, {preserveTimestamps: true})` fails on Windows with error:

>EPERM: operation not permitted, open 'D:\...(dest)'

The error occurs in this part of the code:

Edit: Correction : <strike>https://github.com/jprichardson/node-fs-extra/blame/master/lib/copy/ncp.js#L113</strike>

https://github.com/jprichardson/node-fs-extra/blame/master/lib/copy/ncp.js#L116
```js
        if (preserveTimestamps) {
          utimes.utimesMillis(target, file.atime, file.mtime, function (err) {
            if (err) return onError(err) <<
            return doneOne()
          })
```

I forked the library and these 2 tests are also failing: 

```
2 failing

  1) copy > modification option > when modified option is turned on should have the same timestamps on copy:

      Uncaught AssertionError [ERR_ASSERTION]: 1502926785048 === 1502926785000
      + expected - actual

      -1502926785048
      +1502926785000

      at lib\copy\__tests__\copy-preserve-time.test.js:64:20
      at Array.forEach (<anonymous>)
      at copy (lib\copy\__tests__\copy-preserve-time.test.js:46:17)
      at doneOne (lib\copy\ncp.js:228:16)
      at lib\copy\ncp.js:118:20
      at fs.close.closeErr (lib\util\utimes.js:61:23)
      at node_modules\graceful-fs\graceful-fs.js:43:10
      at FSReqWrap.oncomplete (fs.js:135:15)

  2) copySync > modification option > when modified option is turned on should have the same timestamps on copy:

      AssertionError [ERR_ASSERTION]: 1502926785030 === 1502926785000
      + expected - actual

      -1502926785030
      +1502926785000

      at lib\copy-sync\__tests__\copy-sync-preserve-time.test.js:56:20
      at Array.forEach (<anonymous>)
      at Context.it (lib\copy-sync\__tests__\copy-sync-preserve-time.test.js:40:15)
```

RyanZim:
@laggingreflex Are you on 32-bit windows?

laggingreflex:
64-bit, Window 10

laggingreflex:
Expanding on the `utimes.utimesMillis` the error comes from here: 

https://github.com/jprichardson/node-fs-extra/blame/a061b2507ccdff474a80172c2f5db637bfa53905/lib/util/utimes.js#L55

```js
[...]

PR #479, opened 2017-08-17
fix: windows fs.copy with preserveTimestamps

laggingreflex:
Fixes an issue (#478) on Windows when the destination file had read-only attribute set.

* Do not write {file.mode} when writing (it's done later with fs.chmod anyways)
* Do utimesMillis *before* chmod

### DIFF
--- a/lib/copy/__tests__/copy-preserve-time.test.js
+++ b/lib/copy/__tests__/copy-preserve-time.test.js
@@ -14,16 +14,16 @@ if (process.arch === 'ia32') console.warn('32 bit arch; skipping copy timestamp 
 const describeIf64 = process.arch === 'ia32' ? describe.skip : describe
 
 describeIf64('copy', () => {
-  let TEST_DIR
+  const TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy')
 
   beforeEach(done => {
-    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy')
     require(process.cwd()).emptyDir(TEST_DIR, done)
   })
 
   describe('> modification option', () => {
     const SRC_FIXTURES_DIR = path.join(__dirname, '/fixtures')
     const FILES = ['a-file', path.join('a-folder', 'another-file'), path.join('a-folder', 'another-folder', 'file3')]
+    const readOnlyFile = 'b-file'
 
     describe('> when modified option is turned off', () => {
       it('should have different timestamps on copy', done => {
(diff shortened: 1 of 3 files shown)


lib/copy/ncp.js
```diff
@@ -96,7 +96,7 @@ function ncp (source, dest, options, callback) {
 
   function copyFile (file, target) {
     var readStream = fs.createReadStream(file.name)
-    var writeStream = fs.createWriteStream(target, { mode: file.mode })
+    var writeStream = fs.createWriteStream(target)
```
RyanZim: I'm concerned how this might affect *nix users with a non-standard umask. Is it possible that we could be writing a file with incorrect permissions that wouldn't allow us to do anything to the file? I haven't tested, just wondering.

laggingreflex: Interesting, I'm not sure either. Wouldn't the default permission be such that you'd still be able to do anything to it after writing?

A similar module [cpr](https://github.com/davglass/cpr) has been doing it without it (there's still a [feature-req](https://github.com/davglass/cpr/issues/34) for it).

manidlou:
@laggingreflex thanks for your contribution. We truly appreciate your work. @RyanZim you brought up an interesting point. Please let me dig into this a little! I want to make sure it behaves correctly, especially on Linux with different permissions.

jprichardson:
@laggingreflex thank you. Will you add a test?

lib/copy/__tests__/copy-preserve-time.test.js
```diff
@@ -47,8 +47,34 @@ describeIf64('copy', () => {
           done()
         })
       })
+
+      describe('> with a read-only file', () => {
+        const src = path.join(SRC_FIXTURES_DIR, readOnlyFile)
+        const dest = path.join(TEST_DIR, readOnlyFile)
+
+        beforeEach(() => assertPermission(src, '555', {
+          tryToSet: true,
```
laggingreflex: Would you please take a look at this test I've just added. 

I'm not sure what's going wrong. It works locally on my Windows Linux Bash, which is all I have access to currently. 

I'm setting permissions on a new file `b-file` to read only (444 wasn't working for me (windows bash) so I chose 555). 

I wasn't sure whether git allowed committing files with permissions so I wrote this function to actually set the file permissions in (before) the test itself. 

Travis seems to be complaining that it's setting permission to `644` instead of `555`, and this is the bit I'm not sure why.

lib/copy/__tests__/copy-preserve-time.test.js
```diff
@@ -47,8 +47,34 @@ describeIf64('copy', () => {
...
+            testFile({ preserveTimestamps: true })(readOnlyFile)
+            done()
+          })
+        })
+      })
     })
 
+    function assertPermission (file, permissionString, { tryToSet = false, message } = {}) {
+      const stat = fs.statSync(file)
+      const statPerm = (stat.mode & parseInt('777', 8)).toString(8)
```
laggingreflex: Found this way of converting stats.mode to NNN-type permission string from here: https://github.com/nodejs/node-v0.x-archive/issues/3045

lib/copy/__tests__/copy-preserve-time.test.js
```diff
@@ -47,8 +47,35 @@ describeIf64('copy', () => {
           done()
         })
       })
+
+      describe('> with a read-only file', () => {
+        const src = path.join(SRC_FIXTURES_DIR, readOnlyFile)
+        const dest = path.join(TEST_DIR, readOnlyFile)
+
+        beforeEach(() => assertPermission(src, '444', {
+          tryToSet: true,
```
laggingreflex: Tried with `444` still no luck.

manidlou: Not saying it is the cause of failure for this test, just wondering why using `beforeEach` hook here?!

laggingreflex: Here I'm trying to set the permission of the `src` file to `444` because that's what we need to test, right? - that whether or not those permissions will get copied. 

The same thing is done in [copy-permissions.test](https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy/__tests__/copy-permissions.test.js#L78).

laggingreflex: I think the problem might be due to parent directories of the two files being different - `src` is in fixtures dir and `dest` is in `os.tempdir`. I re-wrote the tests so that it'll first write the `src` file in `tempdir` and then copy it again in the `tempdir` (as `dest`) and now both files have same permissions. It's only when one of the files was in `fixtures` dir that the problem occurred. 

I'm not sure why this is, I've very little knowledge about *nix file permissions. Or maybe I'm doing something else wrong, I don't know.

RyanZim:
Closing this since `copy` has been nearly completely rewritten since this PR was created.

PR #485, opened 2017-09-04
Added '.js' extension to the <main> attribute in package.json

lifaon74:
Added '.js' extension to the `main` attribute in package.json, for commpatibility with various loaders or bundlers (webpack, systemjs, browserify,...) or  any other independent lib which may use package.json.


RyanZim:
Released in 4.0.2 :tada:

PR #491, opened 2017-09-12
Developments for patch release

RyanZim:
This is slated for a patch release, hopefully today, but I want some peer review beforehand.

### DIFF
--- a/lib/json/__tests__/promise-support.test.js
+++ b/lib/json/__tests__/promise-support.test.js
@@ -1,0 +1,38 @@
+'use strict'
+
+const fs = require('fs')
+const os = require('os')
+const fse = require(process.cwd())
+const path = require('path')
+const assert = require('assert')
+
+/* global afterEach, beforeEach, describe, it */
+
+describe('json promise support', () => {
+  let TEST_DIR
+
+  beforeEach(done => {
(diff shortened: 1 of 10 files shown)


lib/json/__tests__/promise-support.test.js
```diff
@@ -0,0 +1,38 @@
...
+  let TEST_DIR
+
+  beforeEach(done => {
+    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'json')
+    fse.emptyDir(TEST_DIR, done)
+  })
+
+  afterEach(done => fse.remove(TEST_DIR, done))
+
+  ;['writeJson', 'writeJSON'].forEach(method => {
```
manidlou: Not a big deal, but I guess the `;` ended up here accidentally?!

RyanZim: No, that's intentional. This is one of the cases where no semicolon code isn't possible. Without the semi, JS ASI will interpret the `[` as an array access on the return value of the `afterEach` call.

Without the semi, the JS engine reads this code as:
```js
afterEach(done => fse.remove(TEST_DIR, done))['writeJson', 'writeJSON'].forEach(...)
```
Which will result in `TypeError: Cannot read property writeJSON of undefined`

Issue #323, opened 2016-12-03
Support for `dest` being a directory when using `copy*()`?


matatk:
When using `copySync()` (and, I assume, `copy()` too): if `src` is a file path and `dest` is a directory path, I would expect that the source file would be copied into the destination directory, with the same name that it had before (i.e. the last part at the end of the `src` path string).

However, it seems that, instead, `copy()` attempts to replace the `dest` directory with the `src` file.

I have tried `{clobber: false}` but this simply seems to stop the copy operation from taking place [as per your explanation in #321 I now understand why].

To get it to behave as desired, I have to include the destination filename at the end of the `dest` path.  This means duplicating the destination filename.  Sometimes, my code that calls `copy()` does not know the filename (only the full `src` path), so it is necessary to make an extra call, to `path.basename()`, first.

I assume that this behaviour is by design to fit in with how (standard) `fs` works? However I wonder if you could add a further `copy()`-like function (or option on the fs-extra version of `copy()` to support this behaviour?

Thanks for this very helpful module (only just started using it, but it’s already saved me lots of time) :-).

jprichardson:
> When using copySync() (and, I assume, copy() too): if src is a file path and dest is a directory path, I would expect that the source file would be copied into the destination directory, with the same name that it had before (i.e. the last part at the end of the src path string).

I understand the perspective here. A lot of people have requested similar functionality. It's my belief that we as programmers should be explicit where possible. As reading code is harder than writing code. `fs-extra` has chosen to encourage this by making programmer intent obvious to those who read the code.

I think where this functionality makes sense is in the case of shell programs where terse expressiveness is valued given the longevity of shell scripts.

Hope this makes sense.

>Thanks for this very helpful module (only just started using it, but it’s already saved me lots of time) :-).

You're welcome. @RyanZim and I have a lot of great things planned for this module! :)

RyanZim:
@jprichardson BTW, we should document this fact clearly somewhere, since a lot of people seem to be asking about this.

matatk:
@jprichardson: thanks for the thoughtful explanation.

I agree with @RyanZim that a note in the docs would help clarify how it works.

Thanks again, both (and for the very quick replies).

just-boris:
>  It's my belief that we as programmers should be explicit where possible.

IMO, the devs should avoid copy-pasting parts as they cause typos and bugs. For example, I want to copy bunch of files


```js
[...]

PR #497, opened 2017-10-05
update copy documentation to show differences with cp

nico29:
A found myself struggling because I was using the `copy` function with file source and directory destination (as you can do using `cp`). The error was related to an `unlink` system call not working.
After some digging I landed on #323 but I found that the docs has not been updated.

So here are a few words about this topic.
Thank you :)


jprichardson:
Thank you @nico29!

Issue #492, opened 2017-09-22
Folders become inoperable after fse.move, because fs.link isn't returning EPERM error [Mac]


Faline10:
**The Issue**
fse.move currently treats files first and folders second. It depends on fs.link to return an EPERM error to check if it's a directory and therefore needs to call moveAcrossDevice() which will eventually call ncp to recursively copy:
https://github.com/jprichardson/node-fs-extra/blob/master/lib/move/index.js#L67

Unfortunately, fs.link doesn't seem to be throwing the expected EPERM error.

OS: MacOS Sierra 10.12.5
Node: 6.11.3

**The Behavior**
This bug is hard to catch because fs.link is still creating the hard link and unlinking the old one. Node documentation doesn't say anything about how it's doing this, it just points to the link command documentation, which specifies that it should throw an EPERM error if you try using it on folders. http://man7.org/linux/man-pages/man2/link.2.html  (The unlink man-pages have a good explanation of EPERM as "The system does not allow unlinking of directories, or unlinking of directories requires privileges that the calling process doesn't have.  (This is the POSIX prescribed error return; as noted above, Linux returns EISDIR for this case.)")
If I try running the link command directly from terminal, the EPERM error is returned.

The problem is that this newly moved folder is then inoperable. I get a 'Finder Wants To Make Changes' pop up asking for my password. (Permissions aren't the issue though, as all the permissions look the same as it would as if created in Finder.) After filling out the pw, the operation still does not occur.
e.g. After calling fse.move(/srcFolder, /destFolder/srcFolder), and then trying to drag  /destFolder/srcFolder back to / in Finder, the pop up occurs. It also occurs if trying to delete the folder. Either way, the fs operation doesn't occur.

Note: We tested this on Windows and did not see the same issue.

**Resolution**
Regardless of what node is doing that causes this issue when using fs.link on a folder, fse.move should not depend only on the EPERM error to be thrown to check if it's a folder, since fs.unlink() may not return the error.

**Workaround**
Now using fs.rename on both files and directories on Mac/Win, which results in one fast fs operation (rename) instead of two (heavy copy + remove).
Another possibility would be to go native with https://github.com/TooTallNate/NodObjC and use NSFileManager to move directories, which would be fast as well, as there is no copy involved if the source and destination are on the same volume.

jprichardson:
[...]

PR #507, opened 2017-11-03
bugfix/492 - Replace fs.link/unlink with fs.rename in fse.move

Faline10:
Fix for: https://github.com/jprichardson/node-fs-extra/issues/492
"Folders become inoperable after fse.move, because fs.link isn't returning EPERM error [Mac]"

Previous move logic:
- If overwriting, use rename
- If not, use fs.link, then fs.unlink. fs.link will throw an EPERM error if it's a directory, in which case, recursively copy.

The downside to the previous logic is that the recursive copy for the folder is a heavy operation, and it took several different fs operations instead of just one. Also, in the case of this bug on Macs, fs.link would partially succeed in moving a folder instead of throwing an error, leading to an inoperable state on the folder.

Updated move logic:
- Use fs.existsSync to prevent overwriting
- Use fs.rename, which should work on both files and folders

Unlike fs.link, fs.rename will overwrite by default, so must do an existence check first to prevent this. I believe this is a decent tradeoff for fixing this bug and avoiding heavy recursive copy operations on folders.

Other Changes:
- Updated tests
- Got rid of ENOTSUP error checking since this isn't returned by rename nor link, and wasn't a unit test case, so not sure what it was doing there: http://man7.org/linux/man-pages/man2/rename.2.html, http://man7.org/linux/man-pages/man2/link.2.html
- Similarly, there was an 'EISDIR' check for fs.link, when that does not appear to be an error thrown by fs.link. There is an 'EISDIR' for rename, though the definition ("newpath is an existing directory, but oldpath is not a directory") does not suit the use case of calling moveAcrossDevice()/recursive copy, so I have removed this check as well.

Ran tests/linter on both Mac and Windows.

### DIFF
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -199,7 +199,7 @@ describe('move', () => {
 
     fse.move(src, dest, err => {
       assert.ifError(err)
-      assert.strictEqual(fs.link.callCount, 1)
+      assert.strictEqual(fs.rename.callCount, 1)
 
       fs.readFile(dest, 'utf8', (err, contents) => {
         const expected = /^sonic the hedgehog\r?\n$/
@@ -238,7 +238,7 @@ describe('move', () => {
 
     fse.move(src, dest, err => {
       assert.ifError(err)
-      assert.strictEqual(fs.link.callCount, 1)
+      assert.strictEqual(fs.rename.callCount, 1)
 
       fs.readFile(dest + '/another-folder/file3', 'utf8', (err, contents) => {
         const expected = /^knuckles\r?\n$/
@@ -284,7 +284,7 @@ describe('move', () => {
 
     fse.move(src, dest, err => {
       assert.ifError(err)
-      assert.strictEqual(fs.link.callCount, 1)
+      assert.strictEqual(fs.rename.callCount, 1)
 
       fs.readFile(dest + '/another-folder/file3', 'utf8', (err, contents) => {
         const expected = /^knuckles\r?\n$/
--- a/lib/move/index.js
+++ b/lib/move/index.js
@@ -66,7 +66,7 @@ function move (src, dest, options, callback) {
           return
         }
 
-        if (err.code === 'EXDEV' || err.code === 'EISDIR' || err.code === 'ENOTSUP') {
+        if (err.code === 'EXDEV' || err.code === 'EISDIR') {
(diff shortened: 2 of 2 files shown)


RyanZim:
It seems we added the `ESDIR` error logic to handle some quirky virtual filesystems: https://github.com/jprichardson/node-fs-extra/commit/24948a4e0ca50b97480664349d6df031b57ddb97 We also have mock tests for this error, which are now failing with your changes. @jprichardson This happened before I was with the project; do you remember any specific details about this? Interested to know your thoughts.

I'm not really comfortable with using a sync method inside an asynchronous function; this could have a significant perf hit.

I also don't like manually constructing fake system-level errors, but if it's a necessary evil, then so be it.

Faline10:
@RyanZim Ah, apologies, I was too excited and forgot to run the tests after the last minute change. Added the EISDIR logic back in. However, the EISDIR logic was previously in the fs.link error handler, so I am not sure if the use case still applies. I am unsure how to manually test the quirky virtual filesystem scenario. Hopefully the test case will suffice. Please let me know if you or JP have any further thoughts here.

I understand your reservations about the sync method. To be clear, since the boolean overwrite check short circuits, it will only run when overwrite is not true. I can try to find a folder with many files and profile move to compare performance in this scenario.

manidlou:
How is this going?! ping @RyanZim @Faline10?

RyanZim:
@manidlou I'm not comfortable merging a PR that adds sync `fs` calls to an async method.

manidlou:
> I'm not comfortable merging a PR that adds sync fs calls to an async method.

@RyanZim I definitely agree with you.

So, @Faline10 will you modify your PR not to use sync function please?

Faline10:
@manidlou @RyanZim This fix is based off the premise of using fs.rename, which is a fix I used for this bug at work. It works in our scenario, but for fs-extra, there needs to be an existsSync check before it because otherwise it will always overwrite. It only does this check when the overwrite option is false, but that could be a majority of the time. It could be that the previous way this worked on folders, through recursively copying, would be just as slow. Due to the holidays, I haven't had time to test this theory. If that sync check doesn't suit this code base, I'm fine with closing this PR and just referencing it as a "work around" solution in the bug, in case anyone encounters the same problem, and leave it up to you as to whether to close the bug or not.

RyanZim:
Superseded by https://github.com/jprichardson/node-fs-extra/pull/549

PR #514, opened 2017-11-09
Add issue template

manidlou:
Added issue template. I am not sure about the wording though! Please change it any way that you think is appropriate! :smile:

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,5 @@
+Please specify the following supporting information if it applies to your issue:
+
+- `Operating System`:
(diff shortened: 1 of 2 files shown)


RyanZim:
How about this:
```markdown
<!-- Please specify the following supporting information if it applies to your issue: -->
- **Operating System:**
- **Node.js version:**
- **`fs-extra` version:**
```
In my experience; the people that tend to file issues without the needed info, and therefore need an issue template, tend to leave the template text almost untouched. Therefore, I recommend putting the instructions in an HTML comment.

PR #517, opened 2017-11-09
Create dest parent after applying filter in copy

manidlou:
When applying filter, I think this is a little bit better than before although it still highly depends on the user's filter conditions!

But at least, if users choose their filter conditions intelligently, this way, `copy()` correctly doesn't create dest parent directory since its child path fails on filter.

This is until @RyanZim adds his changes for filter function.


manidlou:
That's true. Let me see what I can do here!

manidlou:
> Looks good; please squash to one commit.

Sure thing.

manidlou:
Squash done!

PR #518, opened 2017-11-09
Allow copy's filter to return a Promise

RyanZim:


RyanZim:
@manidlou Good point about the perf; wouldn't have much impact if we're just calling a noop function, but with the extra `Promise.resolve` layer, that will probably have a perf hit.

RyanZim:
Updated & fixed merge conflicts.

Issue #464, opened 2017-07-20
copySync empty folder into an existing file


czb:
Copy an empty folder into an existing file. I would expect, an error should be thrown as the file cannot be overwritten by a folder. Since the folder is empty, no error is thrown and no folder is created. If the folder is not empty, `ENOENT` is thrown.

RyanZim:
This behavior is consistent in both `copy` & `copySync`. The behavior does seem a little weird, though. To be honest, this is what I hate about `copy`; there are a million edge cases + always one more that you haven't thought of.

CC @jprichardson @manidlou for thoughts here; so far I haven't pinned down _why_ it's doing this.

RyanZim:
@jprichardson @manidlou ping?

manidlou:
Yeah I also confirm the faulty behavior! However, I applied a possible fix for this issue in #502 that prevents this from happening. When rewriting `copy` is finished, then related changes will be applied to `copySync` as well since they need to be consistent.

manidlou:
Fixed in 5.0.0 :tada:

PR #519, opened 2017-11-11
Rewrite copySync

### LINKED ISSUE #83: Prevent copying a directory into itself

manidlou:
This is the rewrite of `copySync` to match `copy`. It should resolve #83, #198, and #464. Added corresponding tests for those issues. Also, removed the `fixtures` folder that was used by the `copy-sync-preserve-time.test.js`, instead the fixtures directory is created in the code.

Please let me know if anything doesn't look alright!


manidlou:
IDK! it is so strange! the unit test for the case when `preserveTimestamp is true`, sometimes pass and sometimes doesn't! particularly on older node versions! :confused:

manidlou:
Any ideas?

RyanZim:
Will look at timestamps later.

RyanZim:
@manidlou Spent a bunch of time messing with this, can't figure it out. I'm gonna say just put a conditional in place that that flaky test gets skipped on older node versions. @jprichardson Let me know if you disagree here.

manidlou:
> @manidlou Spent a bunch of time messing with this, can't figure it out. I'm gonna say just put a conditional in place that that flaky test gets skipped on older node versions.

I agree. I played with it too, but no success! I am not sure but it seems like it is somehow related to node internals.

manidlou:
If looks good, I will squash this.

manidlou:
@jprichardson do you have any concerns regarding this?

jprichardson:
Nice work @manidlou!

PR #520, opened 2017-11-17
Fixes fs-extra on linux

arcanis:
The `lchown` function is [not available on some systems](https://github.com/nodejs/node/blob/master/lib/fs.js#L1150-L1166). Using `fs-extra` there causes a crash:

    TypeError: Cannot read property 'name' of undefined
      
      at Object.<anonymous>.exports.fromCallback (../node_modules/universalify/index.js:16:26)
      at Object.<anonymous>.api.forEach.method (../node_modules/fs-extra/lib/fs/index.js:50:21)
          at Array.forEach (<anonymous>)

This PR makes sure that all commands exist before trying to convert them.

### DIFF
--- a/lib/fs/index.js
+++ b/lib/fs/index.js
@@ -9,6 +9,7 @@ const api = [
   'chmod',
   'chown',
   'close',
+  'copyFile',
   'fchmod',
   'fchown',
   'fdatasync',
@@ -20,6 +21,7 @@ const api = [
   'link',
   'lstat',
   'mkdir',
+  'mkdtemp',
   'open',
   'readFile',
   'readdir',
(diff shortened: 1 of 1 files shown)


RyanZim:
@arcanis Thanks for the PR. What kind of Linux are you running when you get this error? What filesystem? I'm on Linux myself, but I've never had this problem, though I can see it could happen, looking at the Node.js code.

lib/fs/index.js
```diff
@@ -33,12 +35,13 @@ const api = [
...
-// Add methods that are only in some Node.js versions
-// fs.copyFile was added in Node.js v8.5.0
-typeof fs.copyFile === 'function' && api.push('copyFile')
-// fs.mkdtemp() was added in Node.js v5.10.0
-typeof fs.mkdtemp === 'function' && api.push('mkdtemp')
+].filter(key => {
+  // Some commands are not available on some systems. Ex:
+  // fs.copyFile was added in Node.js v8.5.0
+  // fs.mkdtemp was added in Node.js v5.10.0
+  // fs.lchown is not available on Linux
```
RyanZim: // fs.lchown is not available on some computers

It isn't unavailable on all Linux.

lib/fs/index.js
```diff
@@ -33,12 +35,13 @@ const api = [
...
+  return fs[key]
```
RyanZim: Just for consistency with the previous code, to prevent possible regressions, can we change this to 

    return typeof fs[key] === 'function'

?

RyanZim:
Also, please squash into one commit.

arcanis:
I'm not too sure. I think O_SYMLINK is supposed to only be present on BSD-based systems, at least according to https://github.com/nodejs/node-v0.x-archive/issues/3598 and https://github.com/nodejs/node-v0.x-archive/issues/2142#issuecomment-39622727. Will make the request changes.

RyanZim:
@arcanis Thanks!

PR #541, opened 2018-01-16
fix function isSrcSubdir in the copy-sync.js and copy.js

kolgotko:
The problem occurs when:
`src = /tmp/path/src`
`dest = /prefix/tmp/path/src`


RyanZim:
I'm not sure I fully understand this change; why is the behavior different on Windows?

kolgotko:
Now behavior is tested for Windows too.

manidlou:
Thanks @kolgotko! Will you squash this please?

kolgotko:
That is how it should be?

PR #547, opened 2018-02-02
add fs-extra-debug

jdxcode:
`mfs` uses a very outdated version of `fs-extra` that it requires in `dependencies`. I think it should probably be removed.

My package isn't quite as flexible (yet), but it solves my problem for now.


jprichardson:
Thanks!

Issue #548, opened 2018-02-03
Rewrite move


manidlou:
The current implementation of `move()` is a little buggy. IDK why `fs.link` was used in the first place for moving on the same device as apposed to just using `fs.rename`! @jprichardson and @RyanZim may have better ideas here. Anyway, It seems like using `fs.link` and relying on `EPERM` error is not suitable approach to handle move operation. Please see #526, #492.

So I am not a fan of rewrite but I believe `move()` needs to be reconsidered for improved implementation anyway!

RyanZim:
@manidlou As far as switching away from `fs.link`, I'm fine with that. Do you have other changes in mind, or just that?

manidlou:
@RyanZim that's right. Switching from `fs.link` to `fs.rename` is the main thing I guess. However, AFAIK one of the main differences between `fs.link` and `fs.rename` is in how they behave when dest exists. I tested this one manually to see how `fs.rename` behaves for files and directories!

If we want to use `fs.rename` we need to consider these particularly when `{overwrite: false}` which is our default case: 

_(src and dest are on the same device)_
- when dest is a file that exists:
  - `fs.rename` doesn't throw any errors and replaces the contents of dest with src contents and removes src.

- when dest is a directory that exists:
  - if dest directory is empty, `fs.rename` doesn't throw any errors and moves the contents of src to dest and removes src.
  - if dest directory is not empty, `fs.rename` throws `ENOTEMPTY` error.

For the case of moving across devices, since that's pretty much the copy operation, maybe we can use our own `fs.copy` since `copy` handles both files and dirs itself, so that we don't need to have an extra implementation for copy inside `move` file.

manidlou:
I'll open up a PR shortly to see how we can improve this.

RyanZim:
Fixed in `v6-dev`; will be released later this month.

RyanZim:
@zeroN1 That's intended behavior; fs-extra can't be overwriting a directory with a file.

Issue #526, opened 2017-12-15
move function excute successfully when the overwrite option is false 


jiazhengwen:
- **Operating System:** win7 x64
- **Node.js version:** 7.9.0
- **`fs-extra` version:** 4.0.3

move function excute successfully when the overwrite option is false

RyanZim:
I'm not sure I understand?

manidlou:
yeah @jiazhengwen what do you mean exactly?

jiazhengwen:
@RyanZim @manidlou 

Operating System: win7 x64
Node.js version: 7.9.0
fs - extra version: 4.0.3
function: move

precondition:buld and testbuild folder already exist
code:
fs.move(__dirname + "/build", __dirname + "/testbuild", { overwrite: false }, function(err) {
  console.log(err);
});

expect: Error:EEXIST: file already exists
actual result: move success, testbuild is overwritten.

manidlou:
Aakh! @RyanZim another OS inconsistency! and again another weird behavior (error code) from Windows!

@jiazhengwen what happens is, when `{overwrite: false}` and dest is a directory that already exists, `fs.link` which is internally used by `move()`, on Windows oddly returns `EPERM` error and not `EEXIST` error! :flushed: 

Currently, part of `move()` logic is when `EPERM` error happens, that is most likely a cross-device error and it correctly (as it is documented) performs the move operation. The thing is, Windows is being a bitch here and returns a different error code! which means we have to tweak our implementation to handle OS specific behavior on that.

RyanZim:
Is this related to https://github.com/jprichardson/node-fs-extra/issues/492?

manidlou:
> Is this related to #492?

Kind of. In #492, `EPERM` error not being called. Here, it is being called while not expected to be called!

From all of these, it is obvious to me that we shouldn't rely on `fs.link` and `EPERM` error code! We should come up with more concrete logic for `move()`.

manidlou:
After all of these, I believe `move` needs major reconsideration anyway! I'll open an issue addressing that.

RyanZim:
@manidlou any progress here?

manidlou:
Worked on it a bit, then I got distracted! I'll try to finish it this weekend.

RyanZim:
Fixed in `v6-dev`; will be released later this month.

PR #549, opened 2018-02-04
move: rewrite move to use fs.rename

### LINKED ISSUE #492: Folders become inoperable after fse.move, because fs.link isn't returning EPERM error [Mac]

manidlou:
This is with the hope to improve `move()` to handle cross platform compatibility better!

Should resolve #492, #526, #548.

Since OSes are not consistent in error codes (please see my comments in #548), we either have to handle this when errors happen on `fs.rename` based on the OS that current process running, or check dest before the code reaches `fs.rename`. IDK, in my mind, checking dest seemed easier both in terms of implementation and also easier to understand when you read the code.

I like the functional approach for `isSrcSubdir` that @kolgotko introduced in #541. So added here too :grin:

Please feel free to leave any feedback! If we think it looks good, I'll apply the corresponding changes to `moveSync()` as well.

### DIFF
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -124,6 +124,28 @@ describe('+ move()', () => {
         done()
       })
     })
+
+    it('should overwrite folders across devices', done => {
+      const src = path.join(TEST_DIR, 'a-folder')
+      const dest = path.join(TEST_DIR, 'a-folder-dest')
+
+      fs.mkdirSync(dest)
+
+      setUpMockFs('EXDEV')
+
+      fse.move(src, dest, {overwrite: true}, err => {
+        assert.ifError(err)
+        assert.strictEqual(fs.rename.callCount, 1)
+
+        fs.readFile(path.join(dest, 'another-folder', 'file3'), 'utf8', (err, contents) => {
+          const expected = /^knuckles\r?\n$/
+          assert.ifError(err)
+          assert.ok(contents.match(expected), `${contents} match ${expected}`)
+          tearDownMockFs()
+          done()
+        })
+      })
+    })
   })
 
   describe('> when overwrite = false', () => {
@@ -148,6 +170,16 @@ describe('+ move()', () => {
 
       fse.move(src, dest, err => {
         assert.ifError(err)
+        done()
+      })
+    })
+
+    it('should error if source and destination are the same and source does not exist', done => {
+      const src = path.join(TEST_DIR, 'non-existent')
+      const dest = src
+
+      fse.move(src, dest, err => {
+        assert(err)
         done()
       })
     })
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -69,32 +69,6 @@ describe('+ move()', () => {
       })
     })
 
-    it('should throw error if src is file and dest is directory', done => {
-      const src = path.join(TEST_DIR, 'a-file')
-      const dest = path.join(TEST_DIR, 'a-folder')
-
-      // verify file exists already
-      assert(fs.existsSync(dest))
-
-      fse.move(src, dest, {overwrite: true}, err => {
-        assert.strictEqual(err.message, `Cannot move file '${src}' to directory '${dest}'.`)
-        done()
-      })
-    })
-
-    it('should throw error if src is directory and dest is file', done => {
-      const src = path.join(TEST_DIR, 'a-folder')
-      const dest = path.join(TEST_DIR, 'a-file')
-
-      // verify file exists already
-      assert(fs.existsSync(dest))
-
-      fse.move(src, dest, {overwrite: true}, err => {
-        assert.strictEqual(err.message, `Cannot move directory '${src}' to file '${dest}'.`)
-        done()
-      })
-    })
-
     it('should overwrite the destination directory', done => {
       // Create src
       const src = path.join(TEST_DIR, 'src')
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -5,7 +5,6 @@ const os = require('os')
 const fse = require(process.cwd())
 const path = require('path')
 const assert = require('assert')
-const rimraf = require('rimraf')
 
 /* global afterEach, beforeEach, describe, it */
 
@@ -24,19 +23,16 @@ function createAsyncErrFn (errCode) {
 }
 
 const originalRename = fs.rename
-const originalLink = fs.link
 
 function setUpMockFs (errCode) {
   fs.rename = createAsyncErrFn(errCode)
-  fs.link = createAsyncErrFn(errCode)
 }
 
 function tearDownMockFs () {
   fs.rename = originalRename
-  fs.link = originalLink
 }
 
-describe('move', () => {
+describe('+ move()', () => {
   let TEST_DIR
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


manidlou:
This test never happens in real scenario even when we were using `fs.link` because `fs.link` never throws `EISDIR` error at least according to [link man page](http://man7.org/linux/man-pages/man2/link.2.html), and `fs.rename` ([rename man page](http://man7.org/linux/man-pages/man2/rename.2.html)) throws `EISDIR` when dest is an existing directory but src is not a directory.

```js
it('should move folders across devices with EISDIR error', done => {
    const src = `${TEST_DIR}/a-folder`
    const dest = `${TEST_DIR}/a-folder-dest`

    setUpMockFs('EISDIR')

    fse.move(src, dest, err => {
      assert.ifError(err)
      assert.strictEqual(fs.link.callCount, 1)

      fs.readFile(dest + '/another-folder/file3', 'utf8', (err, contents) => {
        const expected = /^knuckles\r?\n$/
        assert.ifError(err)
        assert.ok(contents.match(expected), `${contents} match ${expected}`)

        tearDownMockFs()

        done()
      })
    })
  })
```

lib/move/__tests__/move.test.js
```diff
@@ -5,7 +5,6 @@ const os = require('os')
 const fse = require(process.cwd())
 const path = require('path')
 const assert = require('assert')
-const rimraf = require('rimraf')
```
RyanZim: Is `rimraf` used elsewhere, or can we remove it as a devDependency?

RyanZim:
Also @manidlou I'd remove the note about the code license; at this point I think it's too divergent to be recognizable.

lib/move/__tests__/move.test.js
manidlou: It is used in a few test files, but we definitely can remove it from our devDependencies since `rimraf` essentially exists within `remove`. In addition, there are 2 more devDependencies that I believe are not needed and can be removed: `read-dir-files` and `secure-random`. We'll take care of them later!

manidlou:
@RyanZim this is the test log when I run new code with old tests. The results were the same on linux and windows.

```bash
Can't write to device. Skipping move test.


  + move() - prevent moving into itself
    > when source is a file
      ✓ should move the file successfully even when dest parent is 'src/dest'
      ✓ should move the file successfully when dest parent is 'src/src_dest'
      ✓ should move the file successfully when dest parent is 'src/dest_src'
      ✓ should move the file successfully when dest parent is 'src/dest/src'
      ✓ should move the file successfully when dest parent is 'srcsrc/dest'
    > when source is a directory
      ✓ should move the directory successfully when dest is 'src_dest'
      ✓ should move the directory successfully when dest is 'src-dest'
      ✓ should move the directory successfully when dest is 'dest_src'
      ✓ should move the directory successfully when dest is 'src_dest/src'
      ✓ should move the directory successfully when dest is 'src-dest/src'
      ✓ should move the directory successfully when dest is 'dest_src/src'
      ✓ should move the directory successfully when dest is 'src_src/dest'
      ✓ should move the directory successfully when dest is 'src-src/dest'
      ✓ should move the directory successfully when dest is 'srcsrc/dest'
      ✓ should move the directory successfully when dest is 'dest/src'
      ✓ should move the directory successfully when dest is very nested that all its parents need to be created
      ✓ should return error when dest is 'src/dest'
      ✓ should return error when dest is 'src/src_dest'
      ✓ should return error when dest is 'src/dest_src'
      ✓ should return error when dest is 'src/dest/src'

  move
    ✓ should rename a file on the same device
    ✓ should not move a file if source and destination are the same
    ✓ should error if source and destination are the same and source does not exist
    ✓ should not move a directory if source and destination are the same
    1) should not overwrite the destination by default
    2) should not overwrite if overwrite = false
    ✓ should overwrite file if overwrite = true
    ✓ should overwrite the destination directory if overwrite = true
    ✓ should create directory structure by default
    ✓ should work across devices
    ✓ should move folders
    3) should move folders across devices with EISDIR error
    ✓ should overwrite folders across devices
    ✓ should move folders across devices with EXDEV error
    clobber
      ✓ should be an alias for overwrite
    > when trying to move a folder into itself
      ✓ should produce an error
    > when actually trying to move a folder across devices
      > just the folder
        - should move the folder


  33 passing (182ms)
  1 pending
  3 failing

  1) move
       should not overwrite the destination by default:

      Uncaught AssertionError [ERR_ASSERTION]: throw EEXIST
      + expected - actual

      -false
      +true
      
      at fse.move.err (lib/move/__tests__/move.test.js:110:14)
      at checkDest (lib/move/index.js:43:39)
      at fs.stat (lib/move/index.js:100:12)
      at node_modules/graceful-fs/polyfills.js:287:18
      at FSReqWrap.oncomplete (fs.js:167:5)

  2) move
       should not overwrite if overwrite = false:

      Uncaught AssertionError [ERR_ASSERTION]: throw EEXIST
      + expected - actual

      -false
      +true
      
      at fse.move.err (lib/move/__tests__/move.test.js:123:14)
      at checkDest (lib/move/index.js:43:39)
      at fs.stat (lib/move/index.js:100:12)
      at node_modules/graceful-fs/polyfills.js:287:18
      at FSReqWrap.oncomplete (fs.js:167:5)

  3) move
       should move folders across devices with EISDIR error:
     Uncaught 
  Error
      at Timeout.setTimeout [as _onTimeout] (lib/move/__tests__/move.test.js:17:19)
```
1 and 2 are not passing because `fs.rename` doesn't throw `EEXIST` error. It was for `fs.link`. In the new code, we handle this by checking the dest beforehand. So, these 2 cases still exist in our tests and they pass there; it's just instead of `EEXIST` error code, we return error with informative message.

About 3, I commented above.

So, my understanding is `move` still maintains its contract. Please correct me if I am wrong.

manidlou:
When I said `fs.rename doesn't throw EEXIST error`, it may throw that error but that's the same as `ENOTEMPTY` error that is thrown when dest is a nonempty directory. That means, `fs.rename` doesn't throw `EEXIST` when dest is an existing file! When dest is an existing file, it still renames it!

We basically care about this when `{overwrite: false}`. When `{overwrite: true}`, dest will be overwritten anyway.

RyanZim:
Just realized I dropped the ball on this one; will try to get some action here.

RyanZim:
JP prefers to release this as a major in case there’s regressions. How do you feel about delaying release until Node 10 is out in April?

RyanZim:
Merged to branch `v6-dev`.

PR #553, opened 2018-02-26
copySync, moveSync: remove extra step in reading loop

wontem:
By some reasons on our environment `fs.readSync` from node 8.x throws error when try to read data from position which more than size of file. Here is the test script with logs in comments:
https://gist.github.com/wontem/da867f66324e1cbe712c717b5fefb944

Since we start using node 8.x and fs-extra 5.x - the issue is disappeared (fs-extra uses `copyFileSync` from node) but with node 8.x and fs-extra 3.0.1 the issue is reproducible.
While investigating the problem I've noticed that the method `copySync` from fs-extra 3.0.1 uses extra step in its read loop. To be more safe I want to suggest remove this step from fallback logic for `copySync` and for `moveSync` in fs-extra 5.x

### DIFF
--- a/lib/copy-sync/copy-sync.js
+++ b/lib/copy-sync/copy-sync.js
@@ -91,11 +91,10 @@ function copyFileFallback (srcStat, src, dest, opts) {
 
   const fdr = fs.openSync(src, 'r')
   const fdw = fs.openSync(dest, 'w', srcStat.mode)
-  let bytesRead = 1
   let pos = 0
 
-  while (bytesRead > 0) {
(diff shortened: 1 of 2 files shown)


RyanZim:
I'm not understanding exactly why these changes are needed and what this fixes?

jprichardson:
Nice work. Is it possible to get a test in place for at least one of these cases? Thanks.

wontem:
@jprichardson Cases I've mentioned are just for environment on our project and they already covered with existed tests. This MR is just for removing extra step in loop.

Issue #552, opened 2018-02-26
why shut my copy task when i wanna to copy a dir with use filter function


moonrailgun:
https://github.com/jprichardson/node-fs-extra/blob/a0d44c1773a72870ecfa1ff378b80a388fb445ef/lib/copy/copy.js#L59-L65

example:
```javascript
fs.copy(srcDir, destDir, {
  filter: () => {
    console.log('run filter')
    return false
  },
}, (err) => {
  if (err) return console.error(err);

  console.log('success!');
});
```
shut down in first file copy!  

i have no idea why about use `cb()` when run `handleFilter `. maybe i just wanna copy a dir and hope to filter some file which i dont want to copy   
is this a reasonable design?

RyanZim:
You're consistently returning `false` from your `filter`, which indicates you don't want to copy anything, so fs-extra just stops.

moonrailgun:
@RyanZim ,i mean although i always return `false`, it should log `run filter` many times if my `srcDir` has many files.  
but just log once and done

manidlou:
@moonrailgun you are strictly returning `false` in your filter function. The way that `copy` handles filter is that if filter function returns `false`, it doesn't do anything. If you want to use a filter function, you essentially need to specify some sorts of conditions. IDK, at least you can set your filter function returns `true` on your src directory, if you wanna see some logs!

moonrailgun:
@manidlou @RyanZim 
Well above code is just a demo. and i write another demo to express my mind.  

now , i have a dir like this:
```
.
├── index.js
└── src
    ├── a
    ├── b
    └── c
```

and the index.js is a test script like this:
```javascript
const fs = require('fs-extra');

fs.copy('./src', './dist', {
  filter: function(src) {
    if(src.indexOf('b') >= 0) return false;//  i dont want to copy file b
    return true;
  }
}, (err) => {
  if (err) return console.error(err);

  console.log('success!');
})
```

Then , i cd to this dir and run `node index.js`  
console output `success!` but the tree is not changed

```
├── index.js
└── src
    ├── a
    ├── b
    └── c
```

and then, i change my `index.js` line 5  
from `if(src.indexOf('b') >= 0) return false;` to `if(src.indexOf('d') >= 0) return false;`(always return true in filter function), everything will be okay.
```
├── dist
│   ├── a
│   ├── b
│   └── c
├── index.js
└── src
    ├── a
    ├── b
    └── c
```

[...]

PR #554, opened 2018-03-02
[BREAKING] copy & copySync: do not overwrite src & dest paths to avoid side effects

manidlou:
Overwriting src and dest paths with their resolved ones would result in some side effects like #552. This is a fix to avoid that.


RyanZim:
@manidlou wouldn't this be technically a breaking change?

manidlou:
> wouldn't this be technically a breaking change?

@RyanZim that's right.

PR #558, opened 2018-03-20
Wrap everything in fs-accepting functions

### LINKED ISSUE #274: Use with memory-fs?
Hi, i love your library, but can you add support for change native fs to https://www.npmjs.com/package/memory-fs or other fs?

masaeedu:
Fixes #274


masaeedu:
Note: lots of leading whitespace added here, please review diff with `-w` (for Github, add [`?w=1`](https://github.com/jprichardson/node-fs-extra/pull/558/files?w=1))

masaeedu:
As expected, we get syntax errors on Node <= 5, but all tests are passing in Node 6 and up now. All that is left to do is find some suitable name under which to export our `prep` function from `index.js`. This function accepts some `fs` module and returns an equivalent with promisified API and extra methods.

masaeedu:
I've gone over and removed the changes that were incompatible with lower Node JS versions. Could someone please take a 👀 at this? @RyanZim @manidlou

RyanZim:
I have looked over this briefly; will have to give this some careful thought. I hope you understand, this is a major change, and needs to be thought out well before merging, so review may take awhile.

masaeedu:
@manidlou What is wrong with that is that you push the need to thread `fs` through everything into the consumer's code, and you can't simply test normal code with a mocked out underlying `fs` implementation.

RyanZim:
My other concern here is how this is going to play with ESM when they come to Node.js.

masaeedu:
@RyanZim That is a concern, I think ESM in Node is going to have some specific way to do mocking that is specific to the Node runtime. It doesn't exist at the moment, it's on their backlog. I think the approach here is still a fairly simple "dumb" way to do things, it doesn't break anything but treeshaking, and that rarely matters in Node JS programs anyway.

PR #566, opened 2018-04-09
Run stricter linter to cleanup problems and force ES6 usage

RyanZim:
I started work to use rest/spread operators, since they're supported in Node 6+. Along the way, I discovered and fixed a few more problems, like regular functions being used instead of arrow functions, missing `'use strict'` statements, using `let` instead of `const`, a `switch` where `if`/`else` does the job fine, etc.

`write()`'s promise implementation in `lib/fs/index.js` required heavy refactoring to avoid using `arguments`, please review this carefully.

I also discovered that we're still using our own ponyfill for `Object.assign()`, despite the fact it's now well supported in Node. So I removed that, and refactored `lib/index.js` while I was at it.


test.js
```diff
@@ -4,11 +4,10 @@ const os = require('os')
 const path = require('path')
 const klaw = require('klaw')
 const Mocha = require('mocha')
-const assign = require('./lib/util/assign')
 
 const argv = require('minimist')(process.argv.slice(2))
 
-const mochaOpts = assign({
+const mochaOpts = Object.assign({
```
JPeer264: How about:

```js
const mochaOpts = {
  ...argv,
  ui: 'bdd',
  reporter: 'dot',
  timeout: 30000
};

Comment (RyanZim) [reply test.js]: @JPeer264 That's object spread properties; which are only supported in Node 8+, so we can't use those yet. Someday, though...

File: lib/fs/index.js [hunk 2]
```diff
@@ -83,23 +84,23 @@ exports.read = function (fd, buffer, offset, length, position, callback) {
...
-  if (typeof arguments[arguments.length - 1] === 'function') {
-    return fs.write(fd, buffer, a, b, c, callback)
+exports.write = function (fd, buffer, ...args) {
+  if (typeof args[args.length - 1] === 'function') {
+    return fs.write(fd, buffer, ...args)
   }
 
   // Check for old, depricated fs.write(fd, string[, position[, encoding]], callback)
   if (typeof buffer === 'string') {
     return new Promise((resolve, reject) => {
```
JPeer264: This here is actually code duplication now, and this `return` is not any different than to the other one. 

I think we can leave that `if`, because here it does not make the difference if there are two or three arguments. Thoughts?

test.js
JPeer264: Oh dang! I guess I am using too much babel in my projects. Good to know though ;)

lib/fs/index.js
RyanZim: You're right; will fix.

PR #568, opened 2018-04-17
copy*() case-insensitive paths

### LINKED ISSUE #565: copy can delete file content
- **Operating System:** Windows
- **Node.js version:** 8.2.1

I realize that fs-extra 5.0.0 uses fs.copy if supported but I believe this bug to be still present and relevant when using node.js versions older than 8.5.
Since the latest stable electron version is still using node.js 8.2.1 I think this might affect a considerable number of applications.

Ok, so there was a patch to prevent copying a file onto itself, but that patch is broken or rather: it doesn't consider all possible cases, mainly: case insensitive file systems.

`if (src === dest) return cb(new Error('Source and destination must not be the same.'))`

will not catch cases like `copy('c:\\dummy.txt', 'C:\\dUmMy.txt')` on a case insensitive drive.
But this is probably not the only possible situation where that check fails, you can easily construct  a case using junction points/directory symlinks  where two paths are different according to a string comparison but are actually the same file.

What happens in this case is that fs-extra creates the destination file empty, overwriting the existing file and then fails because the source file no longer exists for reading.
And to add insult to injury it leaves a file handle open so the empty file can't be removed until the application is closed.

manidlou:
fix #565.

`copy` and `copySync`:

Check paths stricter before performing copy and also check case-insensitive paths. Added tests as well. Also, found some `copySync` test file was actually calling `copy` :grin: (I forgot to change it to sync version after copy-paste it!) Fixed that one as well :smile:

### DIFF
--- a/lib/copy-sync/__tests__/copy-sync-case-insensitive-paths.test.js
+++ b/lib/copy-sync/__tests__/copy-sync-case-insensitive-paths.test.js
@@ -1,0 +1,131 @@
+'use strict'
+
+const assert = require('assert')
+const os = require('os')
+const path = require('path')
+const fs = require(process.cwd())
+
+/* global beforeEach, afterEach, describe, it */
+
+describe('+ copySync() - case insensitive paths', () => {
+  let TEST_DIR = ''
+  let src = ''
+  let dest = ''
+
+  beforeEach(done => {
+    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy-sync-case-insensitive-paths')
+    fs.emptyDir(TEST_DIR, done)
+  })
+
+  afterEach(done => fs.remove(TEST_DIR, done))
+
+  describe('> when src is a directory', () => {
+    it('should behave correctly based on the OS', () => {
+      src = path.join(TEST_DIR, 'srcdir')
+      fs.outputFileSync(path.join(src, 'subdir', 'file.txt'), 'some data')
+      dest = path.join(TEST_DIR, 'srcDir')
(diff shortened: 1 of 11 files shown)


lib/copy-sync/__tests__/copy-sync-case-insensitive-paths.test.js
```diff
@@ -0,0 +1,115 @@
...
+  afterEach(done => fs.remove(TEST_DIR, done))
+
+  describe('> when the source is a directory', () => {
+    it('should behave correctly based on the OS', () => {
+      src = path.join(TEST_DIR, 'srcdir')
+      fs.outputFileSync(path.join(src, 'subdir', 'file.txt'), 'some data')
+      dest = path.join(TEST_DIR, 'srcDir')
+
+      try {
+        fs.copySync(src, dest)
```
JPeer264: What if this function won't throw. The test will pass, as the assertions won't get called, right? Shouldn't we throw any error right after `fs.copySync` so we make sure we trigger the `catch? 

```js
try {
  fs.copySync(src, dest)
  throw new Error('fs.copySync didn\'t throw')
} ...

manidlou: @JPeer264 that's a good point! however, we specifically want to test if `copySync` throws that error. So, I wouldn't throw error manually there, but I think the better way to make sure `copySync` throws that error would be to assert a boolean after try-catch. I'll apply that there.

manidlou:
squashed this :grin:

PR #569, opened 2018-04-18
use strict assert

manidlou:
So,

`const assert = require('assert')` is [deprecated](https://nodejs.org/api/assert.html).

Instead, it is recommended to use `strict` mode:

`const assert = require('assert').strict`.

But, `strict` property is added in `v9.9.0`. Since we still support node >=v6, as a current solution, just changed all of the following deprecated functions to their `strict` functions:

- `assert.equal` -> `assert.strictEqual`
- `assert.notEqual` -> `assert.notStrictEqual`
- `assert.deepEqual` -> `assert.deepStrictEqual`
- `assert.notDeepEqual` -> `assert.notDeepStrictEqual`


manidlou:
squashed this :grin:

PR #578, opened 2018-05-09
Fix fs.promises ExperimentalWarning on Node 10.1.0

### LINKED ISSUE #577: ExperimentalWarning: The fs.promises API is experimental (Node.js 10.1.0)
- **Operating System:** MacOS 10.13.4
- **Node.js version:** 10.1.0

Hello! First of all, thank you for the great work on maintaining fs-extra 🙇 

With Node.js 10.1.0 recently released, we started to see the following warning in our test suite:

```
(node:7885) ExperimentalWarning: The fs.promises API is experimental
```

It turns out the `fs` module provides a new property `promises` which triggers the warning on the first access.

You can try it out yourself:

```
$ node -v
v10.1.0
$ node -e "console.log(typeof fs.promises)"
object
(node:8277) ExperimentalWarning: The fs.promises API is experimental
```

On Node.js version 10.0.0:

```
$ node -e "console.log(typeof fs.promises)"
undefined
```

Can we fix [lib/fs/index.js](https://github.com/jprichardson/node-fs-extra/blob/7d36349abe257c15ac1b216a374de697161ba5f3/lib/fs/index.js) to not emit ExperimentalWarning until the experimental API is actually accessed?

bajtos:
Fixes #577

When fs-extra is loaded on Node.js 10.1.0, an experimental warning is printed:

```
(node:7885) ExperimentalWarning: The fs.promises API is experimental
```

This pull request fixes the problem by defining `fs.promises` as a getter property.

@RyanZim @jprichardson  could you please take a look?

### DIFF
--- a/lib/__tests__/promise.test.js
+++ b/lib/__tests__/promise.test.js
@@ -2,6 +2,8 @@
 
 /* eslint-env mocha */
 
+const assert = require('assert')
+const fs = require('fs')
 const fse = require('..')
 
 const methods = [
@@ -22,4 +24,12 @@ describe('promise support', () => {
       fse[method]().catch(() => done())
     })
   })
+
+  if (Object.getOwnPropertyDescriptor(fs, 'promises')) {
+    it('provides fse.promises API', () => {
+      const desc = Object.getOwnPropertyDescriptor(fse, 'promises')
+      assert.ok(desc)
+      assert.equal(typeof desc.get, 'function')
+    })
+  }
 })
--- a/lib/fs/index.js
+++ b/lib/fs/index.js
@@ -46,6 +46,11 @@ const api = [
 
 // Export all keys:
 Object.keys(fs).forEach(key => {
+  if (key === 'promises') {
+    // fs.promises is a getter property that triggers ExperimentalWarning
+    // Don't re-export it here, the getter is defined in "lib/index.js"
+    return
+  }
   exports[key] = fs[key]
 })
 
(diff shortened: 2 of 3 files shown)


RyanZim:
@JPeer264 Agreed, but we might still need to keep it for quite awhile, especially if all versions of Node 10 have it.

bajtos:
> I think, we should remove that as soon as the warning is gone in the Node core. Thoughts?

As a consumer of fs-extra, I am happy with removing this workaround as soon as the latest Node.js 10.x version does not print the warning.

bajtos:
Thank you for landing and publishing my fix so quickly! ❤️ 🎉

wangjianjun0730:
What dose it mean?And what can I do to remove this warning? The version of my node is v10.1.0.

archervallesfin:
Still having problem in my side still showing the warning.
node

JPeer264:
You are correct `critical` is directly using `6.0.1`. But there is another dependency, `asset-resolver`, which is using the version `5.0.0`, I think that is causing the warning: https://github.com/addyosmani/critical/blob/master/package-lock.json#L203

billtlee:
Hi, I just upgraded to 6.0.1 running node 10.1.0.  The experimental warning went away, but the functions is not doing anything.  For example

const buildPath = path.resolve(__dirname, 'build');
fs.removeSync(buildPath);

is not deleting my build directory.  Is there a work around for now?  I am using this to compile my files.  Thanks!

RyanZim:
@billtlee Please open a new issue with a reduced test case.

Issue #579, opened 2018-05-12
I thought node version 4+ is supported 


omidfi:
Hello, thanks for the nice library :) 
I remember in the ES6 discussions, people mentioned node version 4.9+ for this library. But seems that it actually needs node 6+ because of the spread operators. 

Should that be mentioned in Readme? I'm just hoping it saves some time for someone who needs to support node 4.

RyanZim:
fs-extra v6 dropped Node 4 support since it's EOL: https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md#600--2018-05-01

omidfi:
Thanks. 

Would it help to mention the node version requirement in Readme? 

 Otherwise please feel free to close this.

RyanZim:
I personally hate having to remember to update the README; people can look at the `.travis.yml` and figure it out.

JPeer264:
@RyanZim what about having specified `engines` in the `package.json`? I think that would help more people who are not familiar with a `.travis.yml`.
https://docs.npmjs.com/files/package.json#engines

jprichardson:
@JPeer264 great idea. Would you submit a PR?

PR #580, opened 2018-05-14
Chore: add supported engines

JPeer264:
These engines settings should support Node v6 and v8+

(ref: #579)


jprichardson:
Any reason to explicitly forbid Node v7?

RyanZim:
It's already EOLed, and we don't test on it.

PR #582, opened 2018-05-16
[BREAKING] copy*(): use ino to check identical paths

### LINKED ISSUE #546: CopySync readlinkSync Permission Denied with Electron Asar
- **Operating System:** OSX, El Capitan, High Sierra
- **Node.js version:** v8.1.3

Thanks for all of the great work on the project. It is a core dependency in almost all projects I use.
Recently when upgrading React to v16, we had to upgrade from fs-extra `v3.x`. So we upgraded to v5.0.0 and all appeared to be working well in our Electron app. That is until after we ran from our build. The problem did not happen in any Windows builds, just OSX.

For some reason when the Electron app attempted to copy from the Asar to the fs, it only made it a couple layers deep in the filesystem tree before reporting `permission denied` and `readlink` errors. After much debugging we realized that it was `v5.0.0` and downgraded to `v4.0.3` which met peer dependencies and all worked as it did before.

We were copying files that were in our application's Asar to a sub-directory in the user's path. The directory had a few directories before files were found in them. This didn't pose a problem for the first few parent directories as it got down to the first two layers with proper permissions. The third layer of directories appeared to be created but without proper permissions. Even in Finder, we are unable to open the directories due to permission errors.

The problem doesn't exist using npm start, just running from the output build where it has to copy from the directories in the Asar. It doesn't happen on Windows, just OSX. We tried builds created on different OSX versions and mixing where it was built and where it was installed. We were able to narrow it down to `fs-extra` `v5.0.0` using `copySync`. 

Another oddity, that might not be related at all is that it would randomly display errors on different paths and fail silently on others. It wasn't always consistent on which ones it showed, some of them even didn't get to the point where it created the broken paths, due to the syncronous usage of `copySync`, I thought it would stop execution after one but not a random 2 or 3. That could be our implementation. The other oddity is reporting problems copying a `.DS_Store` file in the path that didn't exist. We tried hunting it down and never could find it. That could be an issue with the build process and most likely not the library, but thought I'd share it in case it helps.

I apologize for not having a small standalone test case to recreate the issue so that it can be addressed. I do have screenshots of the errors when they happened. Even if this issue gets closed due to not providing enough data to recreate (a common open source community practice), this should still be posted for others encountering similar issues and save them time debugging.

screen shot 2018-01-29 at 3 58 43 pm

screen shot 2018-01-29 at 3 45 54 pm

### LINKED ISSUE #565: copy can delete file content

manidlou:
...`copy*()` and long journey of dealing with lots of edge cases and trying to find a proper and yet simple approach to catch as many of those :bug:s as we can!

fix #565. Also possibly fix #546 as we don't use `fs.readlink` anymore to check the paths.

I guess it's breaking as for the case of symlinks, we throw error with descriptive message instead of silently return.

- use `ino` to check if src and dest are the same (Thanks to @TanninOne for bringing that idea). This is beyond checking just paths. Since `fs.lstat` stats the link itself not the actual file that it points to, we have to do an extra `fs.stat` to get the actual `ino` of both src and dest. This essentially relies on node's internal api to give us correct `ino`s! This also catches the cases like when dest parent is a symlink that points to the parent of src! some :astonished: stuff! (we used to only check src and dest themselves not their parents!)

- check `isSrcSubdir` when checking paths before start copying (achieved the same results with shorter code!)

### DIFF
--- a/lib/copy-sync/__tests__/broken-symlink.test.js
+++ b/lib/copy-sync/__tests__/broken-symlink.test.js
@@ -23,9 +23,8 @@ describe('copy-sync / broken symlink', () => {
 
   afterEach(done => fse.remove(TEST_DIR, done))
 
-  it('should copy broken symlinks by default', () => {
-    assert.doesNotThrow(() => copySync(src, out))
-    assert.strictEqual(fs.readlinkSync(path.join(out, 'broken-symlink')), path.join(src, 'does-not-exist'))
+  it('should error if symlink is broken', () => {
+    assert.throws(() => copySync(src, out))
   })
 
   it('should throw an error when dereference=true', () => {
--- a/lib/copy-sync/__tests__/copy-sync-dir.test.js
+++ b/lib/copy-sync/__tests__/copy-sync-dir.test.js
@@ -6,9 +6,9 @@ const path = require('path')
 const assert = require('assert')
 const crypto = require('crypto')
 
-/* global beforeEach, describe, it */
+/* global beforeEach, afterEach, describe, it */
 
-describe('+ copySync()', () => {
+describe('+ copySync() / dir', () => {
   const SIZE = 16 * 64 * 1024 + 7
   let TEST_DIR
   let src, dest
@@ -19,6 +19,8 @@ describe('+ copySync()', () => {
     dest = path.join(TEST_DIR, 'dest')
     fs.emptyDir(TEST_DIR, done)
   })
+
+  afterEach(done => fs.remove(TEST_DIR, done))
 
   describe('> when src is a directory', () => {
     describe('> when dest exists and is a file', () => {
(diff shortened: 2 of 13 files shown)


manidlou:
That's strange! when I rebuild manually on appveyor, all tests pass!

PR #587, opened 2018-05-30
Show support for mode

robertjacobson:
Node fs's `mkdir` (https://nodejs.org/api/fs.html#fs_fs_mkdir_path_mode_callback) supports mode specification. After reviewing the source
`fs-extra` does as well, but is not documented. Update the documentation
to include the `options` parameter and provide a few examples of using
`mode`.

### DIFF
--- a/docs/ensureDir.md
+++ b/docs/ensureDir.md
@@ -1,10 +1,11 @@
-# ensureDir(dir, [callback])
+# ensureDir(dir[,options][,callback])
 
 Ensures that the directory exists. If the directory structure does not exist, it is created. Like `mkdir -p`.
 
 **Aliases:** `mkdirs()`, `mkdirp()`
 
 - `dir` `<String>`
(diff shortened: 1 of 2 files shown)


docs/ensureDir.md
```diff
@@ -13,13 +14,23 @@ Ensures that the directory exists. If the directory structure does not exist, it
...
+}
 
 // With a callback:
 fs.ensureDir(dir, err => {
   console.log(err) // => null
   // dir has now been created, including the directory it is to be placed in
 })
 
+// With a callback and a mode integer
+fs.ensureDir(dir, desireMode, err => {
```
manidlou: should be `desiredMode`.

robertjacobson: Ah, good catch @manidlou. Fixed.

manidlou:
@jprichardson do you have any concerns for this? If not, we can merge it then.

PR #588, opened 2018-05-31
chore: cache node_modules

DanielRuf:


RyanZim:
My only concern here is: what if there's a bug in npm and it doesn't handle a cached node_modules correctly?

DanielRuf:
> My only concern here is: what if there's a bug in npm and it doesn't handle a cached node_modules correctly?

I do not fully understand what you mean.
This is the cache of Travis CI.

RyanZim:
I know it's the Travis/Appveyor cache. Without this, the CI container starts with no node_modules, and npm does a fresh install. With this cache, the CI container already has node_modules from the last install. If there's a bug in npm that only affects fresh installs, but doesn't affect existing installs, we won't notice.

Perhaps I'm over-thinking this, I probably am. Requesting other's opinions here.

DanielRuf:
> If there's a bug in npm that only affects fresh installs

Never had such bugs so far.

DanielRuf:
Also you can always clear the cache in the settings fo the build at Travis.

PR #589, opened 2018-05-31
chore: remove Node.js 9

DanielRuf:


RyanZim:
We still technically support Node 9, so we still need to test it. Sorry.

DanielRuf:
> We still technically support Node 9, so we still need to test it. Sorry.

Well, the odd version numbers are no stable or LTS releases. It is and was always 4, 6, 8 and 10.

Support for 9 has ended like 4 (no bugfixes), 10 replaced 9, see https://github.com/nodejs/Release/raw/master/schedule.png

DanielRuf:
You just need to test 8 and 10 as all bugfixes for 9 will only land in 10.

RyanZim:
Odd numbers are not LTS, but they are stable releases with shorter support timelines since v5. I am aware that support for 9 has ended, but we can't drop support for a Node release without a major version bump. I'm not going to cut a major release _just_ to drop v9; I'll drop it whenever we do the next major release.

DanielRuf:
Ok, understood.

RyanZim:
BTW, I appreciate you trying to help out; I don't want you to feel that I don't want contributions.

PR #592, opened 2018-06-07
missing EEXIST err.code

magicode:


magicode:
Of course, All checks have failed
If you're interested. I'll fix it. like this
https://github.com/jprichardson/node-fs-extra/pull/296

RyanZim:
I was never a huge fan of faking native errors; @manidlou what do you think of this?

manidlou:
> I was never a huge fan of faking native errors

yeah me neither! @magicode will you please tell us why we are doing this? How is it necessary?

magicode:
@manidlou 
To accurately distinguish between types of errors

manidlou:
Then we have to handle all types of errors, `EEXIST` is just one of them. We also have to handle OS inconsistencies in error codes. IDK! I am not sure if I am convinced!

@jprichardson what do you think?

jprichardson:
I have no strong feelings here. I defer to you @RyanZim or @manidlou. I will say that if we need to complicate the code, it better be for a really good reason.

RyanZim:
Closing; faking error codes is a mess, not going there.

magicode:
@RyanZim  OK. revert this https://github.com/jprichardson/node-fs-extra/pull/296/files

RyanZim:
That'd be semver-major, but I'm open to it. @magicode care to make the PR?

magicode:
@RyanZim  Sorry. I thought it was the current code.

PR #609, opened 2018-08-04
[BREAKING] moveSync: refactor to use renameSync

### LINKED ISSUE #608: `moveSync` copies and deletes dirs on external drives
- **Operating System:** macOS 10.13.5 (17F77)
- **Node.js version:** v8.11.2

I'm using `moveSync` to rename a big folder on an external drive. The folder is about 2 GB. It seems like `moveSync` copies the folder then deletes the original, rather than simply renaming the folder, since it takes a few minutes to finish. 

I thought that it was supposed to use [rename](https://github.com/jprichardson/node-fs-extra/pull/549) under the hood, but it seems like that's not the case for me, since when I replace `moveSync` with `renameSync`, the rename is instant.

The path to the folder I'm renaming is `/Volumes/Data/Test/One`, if it matters.

manidlou:
fix #608.

Refactored `moveSync` to be consistent with `move`. Also removed unnecessary `rimraf` package from dev-deps.

### DIFF
--- a/lib/move-sync/__tests__/move-sync-prevent-moving-into-itself.test.js
+++ b/lib/move-sync/__tests__/move-sync-prevent-moving-into-itself.test.js
@@ -182,7 +182,7 @@ function testError (src, dest) {
   try {
     fs.moveSync(src, dest)
   } catch (err) {
-    assert.strictEqual(err.message, `Cannot move '${src}' into itself '${dest}'.`)
+    assert.strictEqual(err.message, `Cannot move '${src}' to a subdirectory of itself, '${dest}'.`)
     assert(fs.existsSync(src))
     assert(!fs.existsSync(dest))
   }
--- a/lib/move-sync/__tests__/move-sync.test.js
+++ b/lib/move-sync/__tests__/move-sync.test.js
@@ -5,7 +5,6 @@ const os = require('os')
 const fse = require(process.cwd())
 const path = require('path')
 const assert = require('assert')
-const rimraf = require('rimraf')
 
 /* global afterEach, beforeEach, describe, it */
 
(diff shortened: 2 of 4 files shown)


manidlou:
@RyanZim I was thinking maybe that'd better we follow your good pattern of creating `v8-dev` branch and include this one as well as #587 to v8 release. Also, I will have more PRs to do some cleanups :grin: tho. So we can include them all to v8 release?

manidlou:
I changed the base to `v8-dev`.

RyanZim:
This will need rebased to fix merge conflicts.

@manidlou you've been very helpful here at fs-extra, and too often, I feel like I'm the bottleneck with the release process. As a result, I've added you to fs-extra on npm (i.e. you now have `npm publish` rights). Of course, as always, we still should get peer review before merging our own PRs. If you have any questions, or need any help with getting a release out, I'm always here, can ask me questions on GH or write me at opensrc@ryanzim.com, whichever you prefer.

manidlou:
@RyanZim thanks a lot for giving me publish rights. I am truly honored to be part of `fs-extra` team! I also appreciate your openness for offering help. Just to confirm with you, when I run `npm owner ls fs-extra`, it only returns you and @jprichardson. IDK if it should include me as well or there is another way of checking this?!

You are right about the conflict. I'll rebase this.

RyanZim:
@manidlou Sorry, seems npm was a little buggy and didn't give you access when I requested it. Now fixed; you should have access.

manidlou:
Thanks @RyanZim! I appreciate it.

manidlou:
@jprichardson are you ok with this?! So that I can merge this since `move*()` still needs some work. I am gonna apply another fix to `move*()` to handle same paths correctly. But this needs to be merged before that one comes in! :grin:

jprichardson:
@manidlou yep, I'm okay with it

PR #611, opened 2018-08-05
fix docs to be consistent with node docs

manidlou:
Node's docs uses the pattern of `,` after `[` for optional parameters. For instance, please take a look at [this one](https://nodejs.org/api/fs.html#fs_fs_appendfilesync_path_data_options). So, figured why not following the same pattern!

### DIFF
--- a/docs/copy.md
+++ b/docs/copy.md
@@ -1,4 +1,4 @@
-# copy(src, dest, [options, callback])
+# copy(src, dest[, options][, callback])
 
 Copy a file or directory. The directory can have contents. Like `cp -r`.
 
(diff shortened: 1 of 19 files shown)


docs/copy.md
```diff
@@ -1,4 +1,4 @@
-# copy(src, dest, [options, callback])
+# copy(src, dest[, options, callback])
```
JPeer264: If we follow the Node pattern we can also add optional options:

```
# copy(src, dest[, options][, callback])

manidlou: @JPeer264 that's a good point! I'll update it then.

manidlou:
@JPeer264 refactored and rebased!

PR #612, opened 2018-08-05
remove*() and empty*(): remove contents as Buffers to handle non-UTF-8 filenames on Linux

### LINKED ISSUE #575: Problems when dealing with invalidly-encoded filenames
- **Operating System:** Debian 9
- **Node.js version:** 8.9.3

Hi there. I ran into some cases where `remove()` was unable to remove a directory due to filename encoding issues. I believe there are similar issues using `empty`, `copy`, and `move` operations (and their sync counterparts - basically anything that relies on `fs.readdir` / `fs.readdirSync`).

My issue arose when trying to `fs.remove()` some directories that were created from an unzip operation. During `remove`s / `rimraf`'s tree walk, some of the returned directories seemed not to exist (although they did), causing the final `unlink` operation to fail (since it wasn't actually successfully emptied).

It seems that, in general, names on a file system are just byte sequences, which are not guaranteed to represent fully valid strings. This causes the bytes-> string -> bytes operation, that happens when listing and then operating on items in a directory using Node, to not always produce the same file name that it read.

This encoding problem has been a known Node issue for a while, which is why an option was added to return `Buffer`s from `fs.readdir`. My suggestion is to update the affected methods to use this `Buffer` option. I'm happy to work on a PR, but I wanted to at least get some feedback and discuss the issue before diving in.

Here are a couple Node issues relating to the file name encoding problem:

https://github.com/nodejs/node-v0.x-archive/issues/2387
https://github.com/nodejs/node/pull/5616

Thanks!

rossj:
This is the first part of fixing issue #575 that deals with non-UTF-8 filenames. I've modified `remove()`, `removeSync()`, `emptyDir()` and `emptyDirSync()` to use `{encoding: 'buffer'}` when calling `fs.readdir*`, in order to more reliably remove contents that may not be convertible to UTF-8. This situation may only be possible on Linux file systems, but the changes and additional unit tests are working on Windows NTFS and Apple APFS.

Further work is required to update `copy*()` (and therefore `move*()`) to close #575, but it was becoming a bit of a rabbit hole, so I wanted to get this part squared away first.

Additionally, as a side effect of these changes, it is also possible to pass a `Buffer` input path to `remove()` and `removeSync()` to remove a directory directly that may have an invalid UTF-8 name. This is not yet possible with the `emptyDir*` functions because that first requires Buffer support in `mkdirs`.

Please advise if you would be willing to review + merge this PR, or if you would like to wait for the additional changes mentioned above.


RyanZim:
@manidlou how do you feel about blocking the v8 release until @rossj can get https://github.com/jprichardson/node-fs-extra/issues/575 completely fixed?

Code here looks good BTW.

manidlou:
@RyanZim yep I am totally fine with that!

@rossj can you please give us a rough estimate of how much work you need to do for this to be complete like when do you think you can have it ready?

Code LGTM too!

manidlou:
@rossj the work on `copy*()` and `move*()` is done on branch `v8-dev`. We'll be waiting for your work to be finished before releasing v8.

rossj:
@manidlou thank you, I should have time to work on this over the weekend

RyanZim:
@rossj ping?

rossj:
@RyanZim sorry for the silence... I kind of went down a rabbit hole and got suck in some upstream issues and scope creap.

For example, #575 is about expanding `Buffer` path input to all fs-extra methods, including `mkdirs`. The problem is that `mkdirs` uses `path.resolve` and `path.normalize` and `path.dirname`, but the `path` module has no equivalents for dealing with Buffers, which lead to nodejs/node#23722, and the conclusion that a `Buffer`-based path module would be a good user-land module. I have slowly been working on one, but it will take some more time.

Furthermore, I discovered nodejs/node#23735 which prevents fs-extra from being able to copy / delete / do anything to certain files on Windows. The main purpose of this PR is to make `remove` & `empty` work with any files on any system, regardless of encoding, but that isn't actually possible at this point (on Windows).

My suggestion is to either leave this Buffer stuff behind and continue to v8 without it, or take the Buffer improvements as they are now with updates only to `remove` and `empty`, which do actually improve / fix some issues on Linux/Unix systems, as fully addressing #575 requires some more free time to implement the buffer-path module.

RyanZim:
I'm leaning towards going ahead with v8 without these changes; @manidlou?

manidlou:
@RyanZim yeah I am fine with releasing v8 without these changes.

RyanZim:
@rossj How's it going?

RyanZim:
Closing this PR as abandoned.

PR #617, opened 2018-08-09
Chore: github issue template

JPeer264:
Seems like couple of issues are due to an old version of `fs-extra`. This might help. I'm not native in english, so the second sentence can be improved indeed (two if's in one sentence? 😅)


JPeer264:
Oh ya that is a great idea. This is just missing a lot of times. Something like:

```
**Code to reproduce:**
```

Any suggestions?

.github/issue_template.md
```diff
@@ -1,4 +1,5 @@
-<!-- Please fill out the following information if it applies to your issue: -->
+<!-- First ensure you installed the latest version of fs-extra -->
+<!-- If your bug still exist please fill out the following information if it applies to your issue: -->
```
RyanZim: `bug still exists`

PR #618, opened 2018-08-12
copy*(): fix copying bind-mounted directories

### LINKED ISSUE #613: copy*() name-based isSrcSubdir check is insufficient, perhaps needs ino
- **Operating System:** Debian Jessie
- **Node.js version:** 10.8

The work that @manidlou did in PR #582 changed a lot of sanity and edge-case checks in `copy()` and `copySync()` to use `ino` instead of name / paths for comparison. However, the `isSrcSubdir()` check still relies on checking string names, which is insufficient in the case that a single folder has multiple locations / names, e.g. a double-mounted Docker volume or a bind-mounted folder. I believe that this check should also be switched to rely on `ino`.

In order to exploit this name check, here is what I did:

1. Made an empty directory on my host Windows OS
2. Started a Docker container, double-mounting the empty dir into `/mount/dir1` and `/mount/dir2`
3. Using `fs-extra` inside the container, run `fs.copySync('/mount/dir1', '/mount/dir2/sub')`. The current `isSrcSubdir()` check will fail to abort the copy, and the operation will recurse infinitely, continuing to create nested `/sub` dirs until execution is stopped manually, or perhaps a stack overflow occurs.

As an additional note, I noticed this issue while working on #575. Removing the string-based checks is a pre-requisite for finishing #575.

manidlou:
fix #613.

Fixed the issue of infinite loop that would used to happen for cases like bind-mounted directories with subdirectory that would end up creating a lot of recursive same subdirectories and finally failing with `ENAMETOOLONG` error.

With this fix, we catch these weird cases and throw error with descriptive message.

Also, cleaned up some tests.

### DIFF
--- a/lib/copy/__tests__/copy-case-insensitive-paths.test.js
+++ b/lib/copy/__tests__/copy-case-insensitive-paths.test.js
@@ -3,7 +3,8 @@
 const assert = require('assert')
 const os = require('os')
 const path = require('path')
-const fs = require(process.cwd())
+const fs = require('../../')
+const platform = os.platform()
 
 /* global beforeEach, afterEach, describe, it */
 
@@ -26,12 +27,12 @@ describe('+ copy() - case insensitive paths', () => {
       dest = path.join(TEST_DIR, 'srcDir')
 
       fs.copy(src, dest, err => {
-        if (os === 'linux') {
+        if (platform === 'linux') {
           assert.ifError(err)
           assert(fs.existsSync(dest))
           assert.strictEqual(fs.readFileSync(path.join(dest, 'subdir', 'file.txt'), 'utf8'), 'some data')
         }
-        if (os === 'darwin' || os === 'win32') {
+        if (platform === 'darwin' || platform === 'win32') {
           assert.strictEqual(err.message, 'Source and destination must not be the same.')
           assert(fs.existsSync(src))
           assert(!fs.existsSync(dest))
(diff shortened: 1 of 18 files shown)


manidlou:
These tests that are failing are not related to these changes and are from `case-insensitive-paths` test. I realized the `case-insensitive-paths` test file was buggy and actually not getting the platform string correctly. So, I fixed that and tests passed on my linux machine, but apparently I have to look at it to see what's the issue with mac and windows!

manidlou:
Cool It is fixed now! :grin:

rossj:
Just some thoughts:

Might it be necessary to check up the whole dest path? For example, in the case of:

```
mkdir -p src/sub1/sub2
ln -s src/ dest
```
```javascript
fs.copy('src', 'dest/sub1/sub2', cb)
```
If so, perhaps the stats / inode information can be cached, to prevent having to re-fetch this information every time a dir in src is copied. Or, perhaps this check only needs to be made for the main src and dest, and not again inside `copyDirItem()` for every subdir in src?

RyanZim:
@manidlou ping?

manidlou:
@RyanZim thanks for pinging! I'll try to finish it this weekend.

manidlou:
It is ready to be reviewed!

lib/copy/__tests__/copy-case-insensitive-paths.test.js
```diff
@@ -3,7 +3,10 @@
 const assert = require('assert')
 const os = require('os')
 const path = require('path')
-const fs = require(process.cwd())
+const fs = require('../../')
+const platform = os.platform()
+
+console.log('platform: ' + platform)
```
JPeer264: Is this kept intentional?

manidlou: Yes @JPeer264 I kept it for debugging purposes just in case!

manidlou:
@jprichardson @JPeer264 do you have any concerns about this?! Because I have the `move*()` PR ready too and that'd be better if we merge this first before that.

jprichardson:
Nice work @manidlou!

Issue #571, opened 2018-04-19
Tests are failing on mac


JPeer264:
- **Operating System:**
macOS 10.13.3
- **Node.js version:**
v6.14.1 - 1 test is failing
v8.11.1 - 3 tests are failing
- **`fs-extra` version:**
`v6-dev` - branch

I don't know why they pass on travis though

---
1 test is the same failing on v6 and v8: `utimes.test.js:31:7`. It asserts `false` but get `true`. The reason might be, that my flash storage is on `APFS` and not on `HFS`. I have no clue how to fix this.

---
The other two in v8 are actually just timestamp checks, which are different for ~4ms:
- `copy-sync-preserve-time.test.js:48:11` 
- `copy-preserve-time.test.js:51:11`

I don't know if that is valid for you, but the assert can be changed to following:
```js
assert.strictEqual(utimes.timeRemoveMillis(toStat.mtime.getTime()), utimes.timeRemoveMillis(fromStat.mtime.getTime()))
```
The above solution can be a bit flaky as well, if the second timestamp is in another second than the other (according to the `timeRemoveMillis` method). E.g. toStat is `1524121168999` and fromStat is `1524121169001` then it fail although there are just 2ms in between.

Or we can check if those two timestamps are in between a threshold of, lets say, 5ms.

RyanZim:
I can reproduce. Seems like Mac can be crazy buggy with timestamps. The CI seems to be running OS X 10.12.6 Sierra, while we're seeing this on High Sierra. Wonder what would happen if we changed the `osx_image`? https://docs.travis-ci.com/user/reference/osx/#OS-X-Version

JPeer264:
Seems like travis won't fail on the newest image: https://travis-ci.org/jprichardson/node-fs-extra/builds/368613666

RyanZim:
Odd. Anyhow, I'm not sure how to best fix this.

JPeer264:
Another solution might also be to use `Promise.all` to call both at the same time

```js
Promise.all([
  fs.statSync(a),
  fs.statSync(b)
]).then((stat) => {
  assert.strictEqual(stat[0].mtime.getTime(), stat[1].mtime.getTime())
})

RyanZim:
If they're sync methods, that won't help anything.

DiegoRBaquero:
Confirmed. Maybe just make sure abs diff is less than 10ms (or an arbitrary threshold)

mbargiel:
[...]

Issue #631, opened 2018-10-11
npm run lint fails with standard-markdown@4.0.2


mbargiel:
**Problem:**
After cloning this repo in order to investigate an Issue and submit a PR (current master head: 287f234817871976f80e28ab50b339471e4c4223) and running `npm install`, I ran into the following problem when I ran `npm run lint`:

```
path.js:28
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received undefined
    at assertPath (path.js:28:11)
    at Object.resolve (path.js:1171:7)
    at Command.<anonymous> (/Users/maximeb/code/playground/node-fs-extra/node_modules/standard-markdown/cli.js:32:28)
    at Command.listener (/Users/maximeb/code/playground/node-fs-extra/node_modules/commander/index.js:315:8)
    at emitNone (events.js:106:13)
    at Command.emit (events.js:208:7)
    at Command.parseArgs (/Users/maximeb/code/playground/node-fs-extra/node_modules/commander/index.js:665:12)
    at Command.parse (/Users/maximeb/code/playground/node-fs-extra/node_modules/commander/index.js:474:21)
    at Object.<anonymous> (/Users/maximeb/code/playground/node-fs-extra/node_modules/standard-markdown/cli.js:40:4)
    at Module._compile (module.js:652:30)
```

**Root cause:**
I tracked this down to a problem in `standard-markdown@4.0.2` which has been fixed in `5.*`. EDIT: I thought `standard-markdown@4.0.1` did not have this issue, but apparently I'm getting the same behavior.

`standard-markdown` issue: https://github.com/zeke/standard-markdown/issues/24
`standard-markdown` fix commit: https://github.com/zeke/standard-markdown/commit/7b83a0e0a93be3a4a9dac3a94521f98bc43f8afb

**Suggested fixes:**
- Implement a workaround in `package.json` as per https://github.com/octokit/rest.js/pull/974/commits/47e23ff03ab68cb016b9d45e1d1674dde4e02227 (*EDIT: This seems to work with 4.0.1 and 4.0.2, potentially an easy fix*)
- Create a `package-lock.json` so that contributors can reproduce your own setup
- Force 4.0.1 as the version for `standard-markdown` (*EDIT: This does not seem to actually fix the issue!*)
[...]

PR #633, opened 2018-10-12
Update dest atime after copy/copy-sync

mbargiel:
This PR provides a fix for the recently failing "copy-preserve-timestamps" (sync & async) tests on MacOS (partial fix for #571). (I believe they have been incorrect for a while but passed by chance due to timing and the coarse 1-s resolution time of HFS+.)

The main observation is that after copying a source file, the `atime` of that file may be modified (probably always will, due to the copy operation implicitly updating the `atime` when it `read`s from the file according to https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_stat_time_values), so if we want to provide a true `preserveTimestamps` option, we need to re-read the source `atime` and use that value when updating the destination timestamps.

### DIFF
--- a/lib/copy-sync/copy-sync.js
+++ b/lib/copy-sync/copy-sync.js
@@ -3,7 +3,7 @@
 const fs = require('graceful-fs')
 const path = require('path')
 const mkdirpSync = require('../mkdirs').mkdirsSync
-const utimesSync = require('../util/utimes.js').utimesMillisSync
+const utimesMillisSync = require('../util/utimes').utimesMillisSync
 const stat = require('../util/stat')
 
 function copySync (src, dest, opts) {
@@ -66,11 +66,13 @@ function mayCopyFile (srcStat, src, dest, opts) {
 function copyFile (srcStat, src, dest, opts) {
   if (typeof fs.copyFileSync === 'function') {
     fs.copyFileSync(src, dest)
-    fs.chmodSync(dest, srcStat.mode)
-    if (opts.preserveTimestamps) {
-      return utimesSync(dest, srcStat.atime, srcStat.mtime)
+    if (opts.preserveTimestamps && (srcStat.mode & 0o200) === 0) {
+      // Make sure the file is writable before setting the timestamp
+      // otherwise openSync fails with EPERM when invoked with 'r+'
+      // (through utimes call)
+      fs.chmodSync(dest, srcStat.mode | 0o200)
     }
-    return
+    return setDestTimestampsAndMode(srcStat, src, dest, opts)
   }
   return copyFileFallback(srcStat, src, dest, opts)
 }
@@ -80,7 +82,7 @@ function copyFileFallback (srcStat, src, dest, opts) {
   const _buff = require('../util/buffer')(BUF_LENGTH)
 
   const fdr = fs.openSync(src, 'r')
-  const fdw = fs.openSync(dest, 'w', srcStat.mode)
+  const fdw = fs.openSync(dest, 'w')
   let pos = 0
 
   while (pos < srcStat.size) {
@@ -89,10 +91,22 @@ function copyFileFallback (srcStat, src, dest, opts) {
     pos += bytesRead
   }
 
-  if (opts.preserveTimestamps) fs.futimesSync(fdw, srcStat.atime, srcStat.mtime)
+  setDestTimestampsAndMode(srcStat, src, fdw, opts)
 
   fs.closeSync(fdr)
   fs.closeSync(fdw)
+}
+
+function setDestTimestampsAndMode (srcStat, src, dest, opts) {
+  const utimesSync = typeof dest === 'string' ? utimesMillisSync : fs.futimesSync
+  const chmodSync = typeof dest === 'string' ? fs.chmodSync : fs.fchmodSync
+  if (opts.preserveTimestamps) {
+    // The initial srcStat.atime cannot be trusted because it is modified by the read(2) system call
+    // (See https://nodejs.org/api/fs.html#fs_stat_time_values)
+    const updatedSrcStat = fs.statSync(src)
+    utimesSync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
+  }
+  chmodSync(dest, srcStat.mode)
 }
 
 function onDir (srcStat, destStat, src, dest, opts) {
--- a/lib/copy/copy.js
+++ b/lib/copy/copy.js
@@ -4,7 +4,7 @@ const fs = require('graceful-fs')
 const path = require('path')
 const mkdirp = require('../mkdirs').mkdirs
 const pathExists = require('../path-exists').pathExists
-const utimes = require('../util/utimes').utimesMillis
+const utimesMillis = require('../util/utimes').utimesMillis
 const stat = require('../util/stat')
 
 function copy (src, dest, opts, cb) {
@@ -95,7 +95,16 @@ function copyFile (srcStat, src, dest, opts, cb) {
   if (typeof fs.copyFile === 'function') {
     return fs.copyFile(src, dest, err => {
       if (err) return cb(err)
-      return setDestModeAndTimestamps(srcStat, dest, opts, cb)
+      if (opts.preserveTimestamps && (srcStat.mode & 0o200) === 0) {
+        // Make sure the file is writable before setting the timestamp
+        // otherwise openSync fails with EPERM when invoked with 'r+'
+        // (through utimes call)
+        return fs.chmod(dest, srcStat.mode | 0o200, (err) => {
+          if (err) return cb(err)
+          return setDestTimestampsAndMode(srcStat, src, dest, opts, cb)
+        })
+      }
+      return setDestTimestampsAndMode(srcStat, src, dest, opts, cb)
     })
   }
   return copyFileFallback(srcStat, src, dest, opts, cb)
@@ -104,21 +113,27 @@ function copyFile (srcStat, src, dest, opts, cb) {
 function copyFileFallback (srcStat, src, dest, opts, cb) {
   const rs = fs.createReadStream(src)
   rs.on('error', err => cb(err)).once('open', () => {
-    const ws = fs.createWriteStream(dest, { mode: srcStat.mode })
+    // explicitly not setting srcStat mode - will do that last
+    const ws = fs.createWriteStream(dest)
     ws.on('error', err => cb(err))
       .on('open', () => rs.pipe(ws))
-      .once('close', () => setDestModeAndTimestamps(srcStat, dest, opts, cb))
+      .once('close', () => setDestTimestampsAndMode(srcStat, src, dest, opts, cb))
   })
 }
 
-function setDestModeAndTimestamps (srcStat, dest, opts, cb) {
-  fs.chmod(dest, srcStat.mode, err => {
-    if (err) return cb(err)
-    if (opts.preserveTimestamps) {
-      return utimes(dest, srcStat.atime, srcStat.mtime, cb)
-    }
-    return cb()
-  })
+function setDestTimestampsAndMode (srcStat, src, dest, opts, cb) {
+  if (opts.preserveTimestamps) {
+    // The initial srcStat.atime cannot be trusted because it is modified by the read(2) system call
+    // (See https://nodejs.org/api/fs.html#fs_stat_time_values)
+    return fs.stat(src, (err, updatedSrcStat) => {
+      if (err) return cb(err)
+      return utimesMillis(dest, updatedSrcStat.atime, updatedSrcStat.mtime, (err2) => {
+        if (err2) return cb(err2)
+        return fs.chmod(dest, srcStat.mode, cb)
+      })
+    })
+  }
+  return fs.chmod(dest, srcStat.mode, cb)
 }
 
 function onDir (srcStat, destStat, src, dest, opts, cb) {
--- a/lib/copy-sync/__tests__/copy-sync-preserve-timestamp.test.js
+++ b/lib/copy-sync/__tests__/copy-sync-preserve-timestamp.test.js
@@ -3,7 +3,8 @@
 const fs = require('../../')
 const os = require('os')
 const path = require('path')
-const utimes = require('../../util/utimes')
+const copySync = require('../copy-sync')
+const utimesSync = require('../../util/utimes').utimesMillisSync
 const assert = require('assert')
 const semver = require('semver')
 const nodeVersion = process.version
@@ -19,21 +20,39 @@ const describeIfPractical = (process.arch === 'ia32' || nodeVersionMajor < 8) ? 
 describeIfPractical('copySync() - preserveTimestamps option', () => {
   let TEST_DIR, SRC, DEST, FILES
 
-  beforeEach(done => {
-    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy-sync-preserve-time')
+  function setupFixture (readonly) {
+    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy-sync-preserve-timestamp')
     SRC = path.join(TEST_DIR, 'src')
     DEST = path.join(TEST_DIR, 'dest')
     FILES = ['a-file', path.join('a-folder', 'another-file'), path.join('a-folder', 'another-folder', 'file3')]
-    FILES.forEach(f => fs.ensureFileSync(path.join(SRC, f)))
-    done()
-  })
+    const timestamp = Date.now() / 1000 - 5
+    FILES.forEach(f => {
+      const filePath = path.join(SRC, f)
+      fs.ensureFileSync(filePath)
+      // rewind timestamps to make sure that coarser OS timestamp resolution
+      // does not alter results
+      utimesSync(filePath, timestamp, timestamp)
+      if (readonly) {
+        fs.chmodSync(filePath, 0o444)
+      }
+    })
+  }
 
   afterEach(done => fs.remove(TEST_DIR, done))
 
   describe('> when preserveTimestamps option is true', () => {
-    it('should have the same timestamps on copy', () => {
-      fs.copySync(SRC, DEST, { preserveTimestamps: true })
-      FILES.forEach(testFile({ preserveTimestamps: true }))
+    ;[
+      { subcase: 'writable', readonly: false },
+      { subcase: 'readonly', readonly: true }
+    ].forEach(params => {
+      describe(`>> with ${params.subcase} source files`, () => {
+        beforeEach(() => setupFixture(params.readonly))
+
+        it('should have the same timestamps on copy', () => {
+          copySync(SRC, DEST, { preserveTimestamps: true })
+          FILES.forEach(testFile({ preserveTimestamps: true }))
+        })
+      })
     })
   })
 
@@ -44,16 +63,11 @@ describeIfPractical('copySync() - preserveTimestamps option', () => {
       const fromStat = fs.statSync(a)
       const toStat = fs.statSync(b)
       if (options.preserveTimestamps) {
-        // https://github.com/nodejs/io.js/issues/2069
-        if (process.platform !== 'win32') {
-          assert.strictEqual(toStat.mtime.getTime(), fromStat.mtime.getTime())
-          assert.strictEqual(toStat.atime.getTime(), fromStat.atime.getTime())
-        } else {
-          assert.strictEqual(toStat.mtime.getTime(), utimes.timeRemoveMillis(fromStat.mtime.getTime()))
-          assert.strictEqual(toStat.atime.getTime(), utimes.timeRemoveMillis(fromStat.atime.getTime()))
-        }
+        // Windows sub-second precision fixed: https://github.com/nodejs/io.js/issues/2069
+        assert.strictEqual(toStat.mtime.getTime(), fromStat.mtime.getTime(), 'different mtime values')
+        assert.strictEqual(toStat.atime.getTime(), fromStat.atime.getTime(), 'different atime values')
       } else {
-        assert.notStrictEqual(toStat.mtime.getTime(), fromStat.mtime.getTime())
+        assert.notStrictEqual(toStat.mtime.getTime(), fromStat.mtime.getTime(), 'same mtime values')
         // the access time might actually be the same, so check only modification time
       }
     }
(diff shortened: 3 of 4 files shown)


lib/copy-sync/__tests__/copy-sync-preserve-time.test.js
```diff
@@ -19,11 +19,18 @@ describeIfPractical('copySync() - preserveTimestamps option', () => {
   let TEST_DIR, SRC, DEST, FILES
 
   beforeEach(done => {
-    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy-sync-preserve-time')
+    TEST_DIR = path.join(os.tmpdir(), 'fs-extra', 'copy-sync-preserve-timestamp')
```
mbargiel: Just making this identical to its async counterpart.

lib/copy-sync/__tests__/copy-sync-preserve-time.test.js
```diff
@@ -19,11 +19,18 @@ describeIfPractical('copySync() - preserveTimestamps option', () => {
...
     SRC = path.join(TEST_DIR, 'src')
     DEST = path.join(TEST_DIR, 'dest')
     FILES = ['a-file', path.join('a-folder', 'another-file'), path.join('a-folder', 'another-folder', 'file3')]
-    FILES.forEach(f => fs.ensureFileSync(path.join(SRC, f)))
+    const timestamp = Date.now() / 1000 - 5
```
mbargiel: If you prefer, I could just hard-code a date in the past rather than use "now minus 5 seconds".

lib/copy-sync/__tests__/copy-sync-preserve-time.test.js
```diff
@@ -43,14 +50,14 @@ describeIfPractical('copySync() - preserveTimestamps option', () => {
...
       if (options.preserveTimestamps) {
-        // https://github.com/nodejs/io.js/issues/2069
-        if (process.platform !== 'win32') {
-          assert.strictEqual(toStat.mtime.getTime(), fromStat.mtime.getTime())
-          assert.strictEqual(toStat.atime.getTime(), fromStat.atime.getTime())
-        } else {
-          assert.strictEqual(toStat.mtime.getTime(), utimes.timeRemoveMillis(fromStat.mtime.getTime()))
-          assert.strictEqual(toStat.atime.getTime(), utimes.timeRemoveMillis(fromStat.atime.getTime()))
-        }
+        // Windows has sub-second support: https://github.com/nodejs/io.js/issues/2069
```
mbargiel: I updated the comment after checking the https://github.com/nodejs/io.js/issues/2069 thread - according to its conclusion, we should no longer need to RemoveMillis for Windows (so maybe I should have switched the conditions around instead, or what...? I'm open to reviewing this section)

lib/copy-sync/copy-sync.js
```diff
@@ -66,7 +66,10 @@ function copyFile (srcStat, src, dest, opts) {
     fs.copyFileSync(src, dest)
     fs.chmodSync(dest, srcStat.mode)
     if (opts.preserveTimestamps) {
-      return utimesSync(dest, srcStat.atime, srcStat.mtime)
+      // Cannot rely on previously fetched srcStat because atime is updated by
+      // the read operation: https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_stat_time_values
+      const updatedSrcStat = fs.statSync(src)
+      fs.utimesSync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
```
mbargiel: I just realized I removed a `return` statement here but it doesn't change the behavior (it was actually unnecessary). Do you want me to re-add it?

mbargiel:
The lint failure is caused by https://github.com/jprichardson/node-fs-extra/issues/631, and fixed in https://github.com/jprichardson/node-fs-extra/pull/636.

mbargiel:
Rebased on top of latest `master` - plus added a fix when the source file is read-only.

lib/copy/copy.js
```diff
@@ -109,18 +109,33 @@ function copyFileFallback (srcStat, src, dest, opts, cb) {
...
 }
 
-function setDestModeAndTimestamps (srcStat, dest, opts, cb) {
-  fs.chmod(dest, srcStat.mode, err => {
+function setDestModeAndTimestamps (srcStat, src, dest, opts, cb) {
+  if (!opts.preserveTimestamps) {
+    return fs.chmod(dest, srcStat.mode, cb)
+  }
+
+  const next = (err) => {
```
mbargiel: This part is a bit tricky. (Check the sync version to better understand.)

Basically, the flow is:
1 -> (optional) set writable if not writable
2 --> re-read atime from source
3 ---> set dest atime 
4 ----> set dest mode

mbargiel: The complexity stems from the optional step. I'm open to suggestions for making this more readable.

RyanZim:
Initial quick review LGTM; but I'd like @manidlou to take a detailed look, since I'm not very familiar with this part of the codebase.

mbargiel:
I checked the coverage loss. 

* In the case of `copySync`, I expanded an else branch in a function that does not seem covered (`copyFileFallback`). (https://coveralls.io/builds/23386203/source?filename=lib%2Fcopy-sync%2Fcopy-sync.js#L101)
* In the case of `copy`, I modified `setDestModeAndTimestamps` and introduced a few additional callback-based operations, and there are no tests that cover the `true` branch if the usual `if (err) return cb(err)` pattern. I would have to set up a couple extra tests where each intermediate step fails, let me know if this is required. (https://coveralls.io/builds/23386203/source?filename=lib%2Fcopy%2Fcopy.js#L112)

lib/copy-sync/copy-sync.js
```diff
@@ -91,7 +98,10 @@ function copyFileFallback (srcStat, src, dest, opts) {
     pos += bytesRead
   }
 
-  if (opts.preserveTimestamps) fs.futimesSync(fdw, srcStat.atime, srcStat.mtime)
+  if (opts.preserveTimestamps) {
+    const updatedSrcStat = fs.statSync(src)
+    fs.futimesSync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
```
manidlou: `fs.futimesSync()` should use `fdw` and not `dest`.

lib/copy/copy.js
```diff
@@ -109,18 +109,33 @@ function copyFileFallback (srcStat, src, dest, opts, cb) {
...
+      return utimes(dest, srcStat.atime, srcStat.mtime, (err) => {
+        if (err) return cb(err)
+        return fs.chmod(dest, srcStat.mode, cb)
+      })
+    })
+  }
+
+  if ((srcStat.mode & 0o200) === 0) {
+    // Make sure the file is writable first
+    return fs.chmod(dest, srcStat.mode | 0o200, next)
```
manidlou: This doesn't seem required! also this is a little hard to read. I think that'd be better we just run the functions without creating another inner function.

lib/copy-sync/copy-sync.js
```diff
@@ -91,7 +98,10 @@ function copyFileFallback (srcStat, src, dest, opts) {
...
+  }
```
manidlou: Also I just realized we forgot to `chmod` again here before closing `fd`s. `copy` does that and we need to be consistent. So I think it is better we have the same function `setDestModeAndTimestamps()` like in `copy` so that we can call it after both native copy and fallback.

lib/copy-sync/copy-sync.js
```diff
@@ -68,10 +68,17 @@ function mayCopyFile (srcStat, src, dest, opts) {
 function copyFile (srcStat, src, dest, opts) {
   if (typeof fs.copyFileSync === 'function') {
     fs.copyFileSync(src, dest)
-    fs.chmodSync(dest, srcStat.mode)
     if (opts.preserveTimestamps) {
-      return utimesSync(dest, srcStat.atime, srcStat.mtime)
+      // Make sure the file is writable first
+      if ((srcStat.mode & 0o200) === 0) {
+        fs.chmodSync(dest, srcStat.mode | 0o200)
```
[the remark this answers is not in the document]
mbargiel: Actually, it is required. One of the tests I added reveals that it doesn't work without this if the source file is read-only (manually tested on Windows and Mac).

1) If `src` is readonly, then so is `dest` after `fs.copyFileSync(src, dest)` 
2) `utimesSync` then fails with a permission error because it's not allowed to open the `dest` file for write.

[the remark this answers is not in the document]
mbargiel: Simple repro on Node 8.11.3, with latest head (9f1c02916bb51d43feb7063f00f138e232f4d518). Assumed to be run from the fse directory.
```
const fse = require('.')
fse.ensureFileSync('src')
fse.chmodSync('src', 0x111)
fse.copySync('src', 'destNoPreserve') // works
fse.copySync('src', 'destPreserve', { preserveTimestamps: true }) // fails with the following stack trace (Windows example):
Error: EPERM: operation not permitted, open 'C:\Users\maximeb\code\node-fs-extra\destPreserve'
    at Object.fs.openSync (fs.js:646:18)
    at utimesMillisSync (C:\Users\maximeb\code\node-fs-extra\lib\util\utimes.js:68:17)
    at copyFile (C:\Users\maximeb\code\node-fs-extra\lib\copy-sync\copy-sync.js:79:7)
    at onFile (C:\Users\maximeb\code\node-fs-extra\lib\copy-sync\copy-sync.js:55:25)
    at getStats (C:\Users\maximeb\code\node-fs-extra\lib\copy-sync\copy-sync.js:50:44)
    at startCopy (C:\Users\maximeb\code\node-fs-extra\lib\copy-sync\copy-sync.js:40:10)
    at handleFilterAndCopy (C:\Users\maximeb\code\node-fs-extra\lib\copy-sync\copy-sync.js:35:10)
    at Object.copySync (C:\Users\maximeb\code\node-fs-extra\lib\copy-sync\copy-sync.js:28:10)
```

With this proposed change, no exception. If I comment out line 74, then I get the same exception.

If you don't want to keep this as-is, I see three options:
- Make sure `copyFileSync` does not set the `dest` mode to be the same as `src`
- Change `utimesMillisSync` to be more robust if it fails with EPERM
- Change `utimesMillisSync` to avoid this problem by not using `fs.openSync(<path>, 'r+')` + `fs.futimesSync` but straight out `fs.utimesSync` which works even on readonly files (note: I'm not sure the `futimes` and `utimes` interfaces are fully compatible, this might be the reason for using `futimes` in `utimesMillisSync` in the first place...)
- Change _this code_ to implement the chmod in a try/catch around utimesSync... but then it's no simpler than an `if`, just more costly.

lib/copy-sync/copy-sync.js
mbargiel: Oops, absolutely!

lib/copy-sync/copy-sync.js
mbargiel: That would make sense. I'll try it out.

lib/copy/copy.js
mbargiel: See `copySync` discussion on the very same topic.

lib/copy-sync/copy-sync.js
mbargiel: Seems like no test verified this code path. Let me check.

mbargiel: Indeed - `copyFileFallback` in both `copy.js` and `copySync.js` is not covered at all, according to coveralls.

mbargiel: Fixed in bd947fdc12c1be6c65206bac45794bbe48d6041f

lib/copy-sync/copy-sync.js
mbargiel: Question, though: if the fallback copy is used, wouldn't it be much better to use the `futimes` and `fchmod` operations directly to make the operation both atomic and faster, rather than closing the file descriptor and calling the path-based `setDestModeAndTimestamps()` function? 

In fact, the `setDestModeAndTimestamps` function could be split in two, one taking paths in, and the other taking file descriptors in. Something like:

- copyFile -> fs.copy -> setDestModeAndTimestamps(src, dest...) -> setDestModeAndTimestampsFd(fsd, fwd...)
- copyFileFallback -> fs.open -> (copy) -> setDestModeAndTimestampsFd(fsd, fwd...) -> fs.close

brodybits:
This fix looks important, what is the status?

lib/copy-sync/copy-sync.js
mbargiel: Not sure the latter works for async since the async copyFileFallback uses streams with implicit FDs. But it seems a good solution for the sync version.

mbargiel: @manidlou What do you think of f55125e? 

I also added tests for copyFileFallback in 1a6eacb - perhaps in a rather unorthodox way, but I quickly figured that we should basically ensure feature parity between both implementations, so why not leverage the same tests...?

mbargiel:
@brodybits I just pushed commits that hopefully address review feedback.

lib/copy/copy.js
mbargiel: Actually I think I managed to make this more readable and less confusing in f55125e. What do you think?

mbargiel:
@manidlou What did you think of the changes I made after your initial review? (I will squash once approved.)

lib/copy-sync/copy-sync.js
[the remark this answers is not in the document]
manidlou: so this is my suggestion:

- do not set the dest mode when we create the dest
- use native `fs.utimes*()` and `fs.futimes*()` (instead of our internal `utimes` implementation)
- if we do the above, then we don't need to check if the file is writable and we can simplify the `setDestTimestampsAndMode` function like this, which is so much easier to read and understand. All tests also pass with this implementation. I tried your example too and that one worked as well:
```js
function setDestTimestampsAndMode (srcStat, src, dest, opts) {
  if (opts.preserveTimestamps) {
    const updatedSrcStat = fs.statSync(src)
    fs.utimesSync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
  }
  fs.chmodSync(dest, srcStat.mode)
}
```
```js
function setDestTimestampsAndModeFd (srcStat, src, fdw, opts) {
  if (opts.preserveTimestamps) {
    const updatedSrcStat = fs.statSync(src)
    fs.futimesSync(fdw, updatedSrcStat.atime, updatedSrcStat.mtime)
  }
  fs.fchmodSync(fdw, srcStat.mode)
}
```

manidlou:
so this is my suggestion:

- do not set the dest mode when we create the dest
- use native `fs.utimes*()` and `fs.futimes*()` (instead of our internal `utimes` implementation)
- if we do the above, then we don't need to check if the file is writable and we can simplify the `setDestTimestampsAndMode` function like this, which is so much easier to read and understand. All tests also pass with this implementation. I tried your example too and that one worked as well:
```js
function setDestTimestampsAndMode (srcStat, src, dest, opts) {
  if (opts.preserveTimestamps) {
    const updatedSrcStat = fs.statSync(src)
    fs.utimesSync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
  }
  fs.chmodSync(dest, srcStat.mode)
}
```
```js
function setDestTimestampsAndModeFd (srcStat, src, fdw, opts) {
  if (opts.preserveTimestamps) {
    const updatedSrcStat = fs.statSync(src)
    fs.futimesSync(fdw, updatedSrcStat.atime, updatedSrcStat.mtime)
  }
  fs.fchmodSync(fdw, srcStat.mode)
}
```

mbargiel:
> so this is my suggestion:
> 
> * do not set the dest mode when we create the dest
> * use native `fs.utimes*()` and `fs.futimes*()` (instead of our internal `utimes` implementation)
> * if we do the above, then we don't need to check if the file is writable and we can simplify the `setDestTimestampsAndMode` function like this, which is so much easier to read and understand. All tests also pass with this implementation. I tried your example too and that one worked as well meaning dests created successfully without any errors.
> 
> ```js
> function setDestTimestampsAndMode (srcStat, src, dest, opts) {
>   const utimesSync = typeof dest === 'string' ? fs.utimesSync : fs.futimesSync
>   const chmodSync = typeof dest === 'string' ? fs.chmodSync : fs.fchmodSync
>   if (opts.preserveTimestamps) {
>     const updatedSrcStat = fs.statSync(src)
>     utimesSync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
>   }
>   chmodSync(dest, srcStat.mode)
> }
> ```


That would certainly be more straightforward. Let me take a stab at that tonight.

mbargiel:
@manidlou I think the problem lies in the following: 
> do not set the dest mode when we create the dest

The code actually relies on `fs.copyFileSync(src, dest)` (see [here](https://github.com/jprichardson/node-fs-extra/pull/633/files#diff-e7565b79a343b45fe91a0d44df7de07aR69)), which will do the full copy, including preserving the mode (though that could be OS-dependent? Not sure - the [nodejs documentation](https://nodejs.org/api/fs.html#fs_fs_copyfilesync_src_dest_flags) on the topic is pretty thin on details). So while in theory I agree with you that your proposed implementation would be more straight-forward, in practice it can only be implemented if we stop relying on `fs.copyFileSync` and implement the copy operation from scratch - in other words, it makes sense in the "copyFileFallback" implementation.

mbargiel:
Or I'm missing something?

mbargiel:
@manidlou https://github.com/jprichardson/node-fs-extra/pull/633/commits/2b7c0e5100647f92f1b0a61d5828c8cb77543725 fixes the tests and now both the "fs.copyFile[Sync]" path and "copyFile[Sync]Fallback" path are tested.

In https://github.com/jprichardson/node-fs-extra/pull/633/commits/fb89a29ecaf87433080559957bfa45c799dac869, I simplified the implementation following your last idea, but I had to keep some of the complexity you didn't like, because of the catch-22 between `fs.utimes` truncating millis and `utils/utimes.utimesMillis` failing if the dest file path is not writable.

mbargiel:
Apparently `atime` does not have the expected value in a bunch of tests... Looking into it.

mbargiel:
I don't understand why these failures are happening. I cannot reproduce them locally, on macOS (10.14), not even with the same node/npm/nvs versions. All these tests pass. The Linux builder also passes.

The failing builds are using macOS 10.13.0, but passing ones seem to be using 10.13.3. Maybe a difference between HFS+ and APS? Could it be there is a race condition between nodejs and the OS APIs, e.g. the source atime is only modified from reading it after the copy is done rather than immediately?

mbargiel:
I locally tested using a HFS+ partition, no luck. This simply isn't happening on macOS 10.14.6.

mbargiel:
@manidlou https://github.com/jprichardson/node-fs-extra/pull/633/commits/f928144b4a48db5c30043fc21f7d6d6212dbcd2b fixes the test failures that I can't quite explain. (Or rather, I can't explain why they were _not_ failures in some builds, considering the Node documentation.) Unfortunately it does so by reviving the "re-read the source file stats" but at least in a way that should be much easier to follow thanks to the suggestions you made in review.

mbargiel:
Note: in https://github.com/jprichardson/node-fs-extra/pull/633#issuecomment-516275474, you suggested not setting the mode after the initial copy so we wouldn't need to check it/make writable before updating the timestamps. Unfortunately, the only way I can think of implementing this is to _not_ call `fs.copyFile`, because it has no option to avoid copying the mode, but instead use what is basically the `copyFileFallback` function. That is: open the streams, copy the file content, set the timestamps, set the mode, close the streams.

mbargiel:
... I don't get it. I locally see the coverage of the `copyFileFallback` paths - why doesn't coveralls see it?

mbargiel:
(Yes, I'll squash it all when it's approved.)

RyanZim:
@manidlou what's the status here?

lib/copy-sync/copy-sync.js
```diff
@@ -89,12 +91,24 @@ function copyFileFallback (srcStat, src, dest, opts) {
...
 }
 
+function setDestTimestampsAndMode (srcStat, src, dest, opts) {
+  const utimesSync = typeof dest === 'string' ? utimesMillisSync : fs.futimesSync
+  const chmodSync = typeof dest === 'string' ? fs.chmodSync : fs.fchmodSync
+  if (opts.preserveTimestamps) {
+    // The initial srcStat.atime cannot be trusted because it is modified by the read(2) system call
+    // (See https://nodejs.org/api/fs.html#fs_stat_time_values)
+    const updatedSrcStat = fs.statSync(src)
+    utimesSync(dest, updatedSrcStat.atime, srcStat.mtime)
```
manidlou: I understand that `srcStat.mtime` and `updatedSrcStat.mtime` should be the same, but having it like this looks a little weird! So why not just using the `updatedSrcStat` for both `atime` and `mtime` to make it easier to read?!

lib/copy/copy.js
```diff
@@ -104,21 +113,27 @@ function copyFile (srcStat, src, dest, opts, cb) {
...
-    }
-    return cb()
-  })
+function setDestTimestampsAndMode (srcStat, src, dest, opts, cb) {
+  if (opts.preserveTimestamps) {
+    // The initial srcStat.atime cannot be trusted because it is modified by the read(2) system call
+    // (See https://nodejs.org/api/fs.html#fs_stat_time_values)
+    return fs.stat(src, (err, updatedSrcStat) => {
+      if (err) return cb(err)
+      return utimesMillis(dest, updatedSrcStat.atime, srcStat.mtime, (err2) => {
```
manidlou: same here as commented above.

lib/copy-sync/copy-sync.js
mbargiel: Fixed in [75756c4](https://github.com/jprichardson/node-fs-extra/pull/633/commits/75756c414d5542cd69789c41ab48dfc6268ae569)

lib/copy/copy.js
mbargiel: Fixed in [75756c4](https://github.com/jprichardson/node-fs-extra/pull/633/commits/75756c414d5542cd69789c41ab48dfc6268ae569)

mbargiel:
Squashed into `3121749c051718887fc66052eeb55f19f2eb309a`. I also reformulated the commit message (but I guess you're free to set whatever merge commit message 😄 )

Glad to have been of any help!

mbargiel:
@RyanZim https://github.com/jprichardson/node-fs-extra/commit/d09b23b32e83fcbf88a2232447ff2aded9a44a26 is an amended version that does not include these fallback tests.

PR #647, opened 2018-11-05
Update graceful-fs version

epicfaace:
This solves an issue with expo not being able to run properly: https://github.com/expo/expo-cli/issues/141

The change from graceful-fs 4.1.2 to 4.1.3 is this commit:
https://github.com/isaacs/node-graceful-fs/commit/a1587eae7c5dcf74ccf6556ac7d6f704ea794804

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "author": "JP Richardson <jprichardson@gmail.com>",
   "license": "MIT",
   "dependencies": {
-    "graceful-fs": "^4.1.2",
+    "graceful-fs": "^4.1.3",
     "jsonfile": "^4.0.0",
     "universalify": "^0.1.0"
   },


RyanZim:
The change that fixes this bug in `graceful-fs` was in `4.1.13`, not `4.1.3`.

`fs-extra` is using `graceful-fs` with a caret version range, so that resolves to the latest version of `graceful-fs` by default. Presumably the issue in expo/expo-cli#141 is caused by a lockfile mandating a specific `graceful-fs` version. Uninstalling and reinstalling `fs-extra` should get the lockfile to pin a newer version of `graceful-fs`.

Closing since `fs-extra`'s version range is correct; this is just a lockfile issue.

epicfaace:
Okay, thanks!

PR #650, opened 2018-11-23
Should have `fs.realpath*.native()` function

gucong3000:
- https://nodejs.org/api/fs.html#fs_fs_realpath_native_path_options_callback


manidlou:
That's right we have to add that! but your PR doesn't add it to the api. You gotta add it to the `api` array in `index.js`.

PR #659, opened 2019-02-11
Clarify `remove` semantics

wmertens:


manidlou:
@wmertens what is the issue exactly? what is confusing?

wmertens:
@manidlou see the patch, it was not clear from the documentation how the call behaves when the file is missing or what it returns

RyanZim:
The distinction between `null` and `undefined` for the return value is untested, and shouldn't be relied on. I would have assumed that

> Like `rm -rf`.

would have conveyed the message that there's no error if the file is nonexistent, but I suppose it wouldn't hurt to explicitly mention that.

manidlou:
@wmertens I agree with @RyanZim that we need to test the `null` and `undefined` value if we want to add it to the docs.

> would have conveyed the message that there's no error if the file is nonexistent, but I suppose it wouldn't hurt to explicitly mention that. :+1: 

I agree with that too! I believe that's a good idea to explicitly mention that in the docs.

Issue #461, opened 2017-07-13
Copy Errors should have code property like Node


Faline10:
Requested enhancement: fse is a stand-in for fs and as such you would expect similar errors, including the code property. One should be able to check `error.code ===' EEXIST'`, for example. Instead we are checking `err.message === 'Source and destination must not be the same.'`

Update: This is specifically for calls to copy. I see that move, for instance, has error codes.

Node reference: https://nodejs.org/api/errors.html#errors_node_js_error_codes

RyanZim:
PR welcome to add any error codes where it makes sense; however, src and dest being the same for a copy operation isn't really an `EEXIST` error.

RyanZim:
Closing for lack of action.

PR #661, opened 2019-03-04
Provide 'EEXIST' code for "already exists" errors

urugator:
So we don't have to match against the error message.

Related: https://github.com/jprichardson/node-fs-extra/issues/461

### DIFF
--- a/lib/copy-sync/copy-sync.js
+++ b/lib/copy-sync/copy-sync.js
@@ -57,7 +57,9 @@ function mayCopyFile (srcStat, src, dest, opts) {
     fs.unlinkSync(dest)
     return copyFile(srcStat, src, dest, opts)
   } else if (opts.errorOnExist) {
-    throw new Error(`'${dest}' already exists`)
+    const existsError = new Error(`'${dest}' already exists`)
+    existsError.code = 'EEXIST'
+    throw new Error(existsError)
   }
 }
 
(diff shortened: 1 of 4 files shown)


RyanZim:
Closing for the same reasons as https://github.com/jprichardson/node-fs-extra/pull/592.

urugator:
Can't we pass `fs.constants.COPYFILE_EXCL` flag to `fs.copyFile` when `overwrite === true` and catch/rethrow the native error based on `errorOnExist`?
It would: 
- make the existence check atomic (current check isn't safe in concurrent env) 
- provide a way to handle the error meaningfully

RyanZim:
The problem with that is that we still need to support environments that don't support `fs.copyFile`.

urugator:
The fallback uses `createWriteStream`:

> The exclusive flag 'x' (O_EXCL flag in open(2)) ensures that path is newly created. On POSIX systems, path is considered to exist even if it is a symlink to a non-existent file. The exclusive flag may or may not work with network file systems.

Is that viable?

What is the suggested way to handle `errorOnExist` error?
I mean, the existence check isn't reliable enough as a guarantee for data loss prevention.
But it's also hard to use as a naive "let the user know" check, because the error is "indistinguishable" from others.    
So what's the use case? Should the `overwrite` and `errorOnExists` option even exist?

Issue #657, opened 2019-01-15
copy and copySync can fail with 'Source and destination must not be the same'


bvachon:
- **Operating System: Windows 10**
- **Node.js version: 8.11.3**
- **`fs-extra` version:7.0.0**

copy and copySync will fail with 'Source and destination must not be the same' when node occasionally returns the same ino for different files due to a bigint rounding issue (https://github.com/nodejs/node/issues/12115). The solution is probably to set options parameter to { bigint: true } when calling fstat (https://nodejs.org/api/fs.html#fs_fs_fstat_fd_options_callback) from the checkPaths function of copy.js and copy-sync.js.

manidlou:
yeah but we should note that `bigint` option was added in `v10.5.0` and we are still supporting node `>=6`.

oliversalzburg:
This is pretty bad and makes tools that depend on this library very unreliable :( As a consumer, is there anything that can be done to work around this issue?

manidlou:
It is unfortunate that older versions of node are buggy in terms of inodes! I agree we need to fix this!

oliversalzburg:
Can you elaborate on the intention of the check?

My thoughts were, if it was intentional to not just check source against destination filename, the intention was to not copy files where the source and destination are some sort of link from one to the other. In which case a copy of a link onto its target or vice versa should be prevented.

I'm not sure what the outcome would be of such a copy operation, but I guess it's undesirable or will fail.

Sadly, in my test case, the entire `fs.Stats` object seems to be identical for both files, as they are actually copies of the same file. So none of the values in that object could be utilized to determine if the files are actually the same or not.

In case the check is in place to prevent failures, I think the only solution is to just let the process fail, catch the error and handle it appropriately.

manidlou:
@oliversalzburg, we mainly check paths to ensure `copy` doesn't end up with unintended results! We used to check just paths as strings but as you mentioned, obviously that is not enough to handle cases like symlinks.
Basically, implementing cross-platform `copy` that handles all types of files, dirs, and links in one shot is not easy. There are a lot of edge cases when it comes to `copy`. We've been trying really hard to abstract all of these away to provide an easy to use api so that the end users don't need to deal with any of that!
So in terms of solution, we can do couple of things here. @jprichardson @RyanZim @JPeer264 I'd like to know your thoughts on this too please!

1. add a new option to copy, something like `checkPaths: true|false`, and let users decide if they want us to check the paths before copying or not.
[...]

PR #666, opened 2019-03-27
CheckPath Error -> Warn of Inode.

Domvel:
Fixes https://github.com/jprichardson/node-fs-extra/issues/657

The inode number is an integer unique to the volume upon which it is stored.
That means, the method checkPaths() disallows me to copy a file from one to another volume.
Also `fs-extra` does not set the big-integer flag of node `fs`. It's only a "default" JavaScript number (max integer is 53-bit). [See the difference here](https://stackblitz.com/edit/js-bigint-comparison?embed=1&file=index.js) On Windows the cap is reached.

No option provided, because checking if Inode is the wrong way. It's not only a big-integer issue. Inodes are only unique to the volume upon which it is stored.

This will fix the issue #657 
I keep it as warning to see it when is happen.
Maybe not the best solution. But with the thrown error, it is not possible to use it.
Example: It's not possible to stable build with Android-Cordova 8.x. Because Cordova switched to `fs-extra` since version 8.x. 

Better solutions are welcome. But keep in mind that the Inode (state.ino of fs) is not really unique global and in current fs-extra implementation. Again:
- The inode number is an integer unique to the volume upon which it is stored.
- `state.ino` of `fs` also supports `big-integer`. Node version 10.4 is required to use this. But currently it's set for `number` (53-bit) only.

The aim is to remove / replace this check. Because it's faulty. (Inode) You'll get errors from `fs` / `os`.

### DIFF
--- a/lib/copy-sync/copy-sync.js
+++ b/lib/copy-sync/copy-sync.js
@@ -182,7 +182,11 @@ function checkStats (src, dest) {
 function checkPaths (src, dest) {
   const {srcStat, destStat} = checkStats(src, dest)
   if (destStat.ino && destStat.ino === srcStat.ino) {
-    throw new Error('Source and destination must not be the same.')
+    console.warn(
+      'Source and destination must not be the same.',
+      'Dest Inode: ' + destStat.ino,
+      'Src Inode: ' + srcStat.ino
+    )
   }
   if (srcStat.isDirectory() && isSrcSubdir(src, dest)) {
     throw new Error(`Cannot copy '${src}' to a subdirectory of itself, '${dest}'.`)
--- a/lib/copy/copy.js
+++ b/lib/copy/copy.js
@@ -234,7 +234,11 @@ function checkPaths (src, dest, cb) {
     if (err) return cb(err)
     const {srcStat, destStat} = stats
     if (destStat.ino && destStat.ino === srcStat.ino) {
-      return cb(new Error('Source and destination must not be the same.'))
+      console.warn(
+        'Source and destination must not be the same.',
+        'Dest Inode: ' + destStat.ino,
+        'Src Inode: ' + srcStat.ino
+      )
     }
     if (srcStat.isDirectory() && isSrcSubdir(src, dest)) {
       return cb(new Error(`Cannot copy '${src}' to a subdirectory of itself, '${dest}'.`))


iamsim:
@Domvel, Ive been following you in all the comments from various bugs referring the same issue.
I just want to ask one thing, have you found any work around for this?
I also cannot build the app since 2 days and am in trouble. Please help me!

Domvel:
@iamsim Oh, I have a stalker. 😄 There is no software-workaround. Because the method `checkPaths` is fixed invoked by `fs-extra`. And Cordova just use it. We built up a linux server for jenkins to build it. 

Until this bug will not be fixed, there are two known "workarounds" It's more a dilemma:
- Build on Linux and keep Cordova `8.x`.
- Downgrade to Cordova `7.x` and also build on Windows.

If the maintainers of `fs-extra` will fix the bug by allow big-integers, the bug of copy between different volumes are still present. If they still comparing the Inode (state.ino of fs).
Or Cordova should replace `fs-extra` again. 

As I suggested: Just remove the state.ino check and let the OS or the underlayed libs throw errors, if source === destination. 
But for you, you should build on linux or downgrade.

iamsim:
> Downgrade to Cordova 7.x and also build on Windows. Because Cordova 7.x does not use fs-extra.

Doing this now!
Will keep you updated if I find a fix.
But please keep me in mind if you find a fix.

Really thank you for a response. 🙂

lib/copy-sync/copy-sync.js
```diff
@@ -182,7 +182,11 @@ function checkStats (src, dest) {
 function checkPaths (src, dest) {
   const {srcStat, destStat} = checkStats(src, dest)
   if (destStat.ino && destStat.ino === srcStat.ino) {
-    throw new Error('Source and destination must not be the same.')
+    console.warn(
```
bvachon: Considering your comment on the fact that inodes can be the same when copying across volumes, then this is a totally valid use case, I would remove this warning and the if statement alltogether.

Domvel: This is just a hotfix to disable the Inode check completely and just log a warning. Not only when copying across volumes. Anyway, this pull-requests can be closed. If I understand it correctly, the issue will be fixed in #667 too.

manidlou:
Closing this as the issue is fixed in `fs-extra` v8.

PR #682, opened 2019-05-13
Add support for fs.realpath.native in envs that support it

RyanZim:
Closes https://github.com/jprichardson/node-fs-extra/pull/650

### DIFF
--- a/lib/fs/__tests__/realpath.test.js
+++ b/lib/fs/__tests__/realpath.test.js
@@ -1,0 +1,24 @@
+'use strict'
+
+const fs = require('fs')
+const fse = require('../..')
+const assert = require('assert')
+
+/* eslint-env mocha */
+
+// fs.realpath.native only available in Node v9.2+
+if (typeof fs.realpath.native === 'function') {
+  describe('realpath.native', () => {
+    it('works with callbacks', () => {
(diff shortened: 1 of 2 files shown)


manidlou:
good! so we don't need to do anything for `fs.realpathSync.native`?

RyanZim:
I think that should work without anything special, since we don't have promises in the mix, but it wouldn't hurt to add a test.

lib/fs/__tests__/realpath.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+/* eslint-env mocha */
+
+// fs.realpath.native only available in Node v9.2+
+if (typeof fs.realpath.native === 'function') {
+  describe('realpath.native', () => {
+    it('works with callbacks', () => {
+      fse.realpath.native(__dirname, (err, path) => {
+        assert.ifError(err)
+        assert.equal(path, __dirname)
```
manidlou: It is better to use `assert.strictEqual()` since `assert.equal()` is deprecated in the legacy mode.

manidlou: It is better to use `assert.strictEqual()` since `assert.equal()` is deprecated in the legacy mode.

PR #694, opened 2019-06-01
Add heuristics to stats.ino comparison when bigint is not available

mbargiel:
The resolution of https://github.com/jprichardson/node-fs-extra/issues/657 relies on Node >= 10.5, but actually leaves the issue not fixed on previous versions of Node (in particular, for the OP, who reported the issue on Node 8.11.3). The issue particularly problematic and frequent in automated setups, such as running tests that create files in sequence, where it is likely to happen.

This PR minimizes the frequency of this issue by comparing additional `fs.Stats` values when `Bigint` support is not available. Note that the check is still heuristic: different stats prove they are different files*, but while same stats do not prove they are the same, we must err on the side of caution and still treat them as the same file.

(*) **Race condition**: comparing file stats before performing a copy or other operation implicitly introduces a race condition. However the existing implementation already suffered from this, so we presume that this PR does not make the condition much worse (with the exception that unlike `ino` and `vol`, a malicious user could in theory manipulate some of the other `stats`). In practice, this race condition seems extremely hard to abuse, since it requires using a vulnerable system, and a means to change the file stats between the two `stats` calls made by `getStats` and `getStatsSync`.

### DIFF
--- a/lib/util/stat.js
+++ b/lib/util/stat.js
@@ -76,7 +76,7 @@ function checkPaths (src, dest, funcName, cb) {
   getStats(src, dest, (err, stats) => {
     if (err) return cb(err)
     const { srcStat, destStat } = stats
-    if (destStat && destStat.ino && destStat.dev && destStat.ino === srcStat.ino && destStat.dev === srcStat.dev) {
+    if (destStat && areIdentical(srcStat, destStat)) {
       return cb(new Error('Source and destination must not be the same.'))
     }
     if (srcStat.isDirectory() && isSrcSubdir(src, dest)) {
@@ -88,7 +88,7 @@ function checkPaths (src, dest, funcName, cb) {
 
 function checkPathsSync (src, dest, funcName) {
   const { srcStat, destStat } = getStatsSync(src, dest)
-  if (destStat && destStat.ino && destStat.dev && destStat.ino === srcStat.ino && destStat.dev === srcStat.dev) {
+  if (destStat && areIdentical(srcStat, destStat)) {
     throw new Error('Source and destination must not be the same.')
   }
   if (srcStat.isDirectory() && isSrcSubdir(src, dest)) {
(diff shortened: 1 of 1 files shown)


mbargiel:
Hmm apparently tests fail on CI (but not locally, at least not on Mac). Looking into it...

mbargiel:
Sorry about the last commit (10b96d3), I pushed too fast - it's unnecessary. I removed it.

mbargiel:
Checked the coverage loss - it's due to the fact that the code branch where I added the heuristic checks was not covered by tests (of course, since BigInt is supported in the test runs used by the coverage check).

lib/util/stat.js
```diff
@@ -147,12 +147,35 @@ function checkParentPathsSync (src, srcStat, dest, funcName) {
...
     throw err
   }
-  if (destStat.ino && destStat.dev && destStat.ino === srcStat.ino && destStat.dev === srcStat.dev) {
+  if (areApparentlyIdentical(srcStat, destStat)) {
     throw new Error(errMsg(src, dest, funcName))
   }
   return checkParentPathsSync(src, srcStat, destParent, funcName)
 }
 
+function areApparentlyIdentical (srcStat, destStat) {
```
manidlou: This is good but the function name is a bit confusing. Will you please rename it to something simpler like just `areIdentical`?

mbargiel: Sure thing

mbargiel: Fixed in https://github.com/jprichardson/node-fs-extra/pull/694/commits/7e4c8525ea031d89cd8e7d40245afc3b753c7559

mbargiel:
@manidlou Commits squashed into 391fb7365e8fa53af5eadf8d673eda1fdda30db9

manidlou:
@RyanZim do we wanna merge this directly into master? we have #633 and #698 that are apparently gonna be ready soon as well!

RyanZim:
I'm good with merging.

PR #698, opened 2019-06-20
Consistent errors across OSs for ensurefile

### LINKED ISSUE #696: ensureFile misleading error message
- **Operating System:** Windows 10
- **Node.js version:** 10.11.0

Say you have a *file* /foo/bar.txt
and call ensureFile('/foo/bar.txt/narf.txt') the error message you get is ENOENT for the path /foo/bar.txt/narf.txt which is confusing, because the purpose of the call is to create the file, why is it failing with the file being missing?

The cause is that pathExists (or fs.exists) only tests for the existence of the filesystem object, it doesn't verify whether it's a directory, so in this case the function proceeds to try to create /foo/bar.txt/narf.txt even though bar.txt is not a directory.

I think instead the error message should be EEXIST with path /foo/bar.txt - signaling that there is a file "bar.txt" that can't be replaced with the required intermediate directory required.

I do realize this is a minor thing but I feel like good error messages are important to quickly identify issues.

lukechilds:
Resolves #696 

Ensures that if you do something like `ensureFile('/foo/bar.txt/narf.txt')` an `ENOTDIR` error is always returned.

Before this PR `ENOTDIR` is returned for macOS and Linux but Windows returns `ENOENT` which is misleading.

### DIFF
--- a/lib/ensure/file.js
+++ b/lib/ensure/file.js
@@ -8,9 +8,21 @@ const pathExists = require('../path-exists').pathExists
 
 function createFile (file, callback) {
   function makeFile () {
-    fs.writeFile(file, '', err => {
+    const dir = path.dirname(file)
+
+    fs.stat(dir, (err, stats) => {
       if (err) return callback(err)
-      callback()
+      if (!stats.isDirectory()) {
+        // This is just to cause an internal ENOTDIR error to be thrown
+        fs.readdir(dir, err => {
+          if (err) return callback(err)
+        })
+      } else {
+        fs.writeFile(file, '', err => {
+          if (err) return callback(err)
+          callback()
+        })
+      }
     })
   }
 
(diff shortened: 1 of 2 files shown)


lukechilds:
@RyanZim I'm calling `fs.readdirSync(dir)` just to force trigger an internal `ENOTDIR` error. What do you think about this?

Seems a bit hacky, but I don't think there's a way to manually create internal Node.js errors?

Alternatively we could just return our own custom error for all OSs.

lib/ensure/file.js
```diff
@@ -8,9 +8,21 @@ const pathExists = require('../path-exists').pathExists
...
+        // This is just to cause an internal ENOTDIR error to be thrown
+        fs.readdir(dir, err => {
+          if (err) return callback(err)
+        })
+      } else {
+        fs.writeFile(file, '', err => {
+          if (err) return callback(err)
+          callback()
+        })
+      }
```
lukechilds: Not too happy with this, seems a bit unreadable. It's been a while since I've worked with callbacks. Cleanup suggestions welcome.

lukechilds:
@RyanZim This PR resolves the related issue.

Let me know if you're ok with the method used.

PR #720, opened 2019-09-06
Upgrade 'standard' and run 'standard --fix'

abetomo:


lib/copy-sync/__tests__/copy-sync-prevent-copying-into-itself.test.js
```diff
@@ -60,59 +60,59 @@ describe('+ copySync() - prevent copying into itself', () => {
...
         const dest = path.join(TEST_DIR, src.replace(/^\w:/, ''))
         return testSuccess(src, dest)
       })
-      it(`of itself`, () => {
+      it('of itself', () => {
         const dest = path.join(src, 'dest')
         return testError(src, dest)
       })
-      it(`should copy the directory successfully when dest is 'src_dest'`, () => {
+      it('should copy the directory successfully when dest is \'src_dest\'', () => {
```
RyanZim: Does the new standard version allow this?

```suggestion
      it("should copy the directory successfully when dest is 'src_dest'", () => {
```

abetomo: Thank you for the review.
No error with double quotes.
it changed.

RyanZim: OK, I realize this is a ton of work, but there's quite a few instances of this. I very much dislike backslash-escaped quotes `\'`, if we could use double-quotes there instead? If you don't want to do it, I'll get to it sooner or later.

abetomo: Changed to double quotes.

PR #724, opened 2019-09-21
Working with Multiple Files and Dirs

EsTharian:
multi and multiSync methods for working with multiple files and directories.

Also, their docs in [multi.md](https://github.com/EsTharian/node-fs-extra/blob/master/docs/multi.md) and [multi-sync.md](https://github.com/EsTharian/node-fs-extra/blob/master/docs/multi.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,21 +110,20 @@ Examples:
 ```js
 const fs = require('fs-extra')
 
-let fileOptions = {
+let files = {
   '/tmp/myfile': {
-    target: '/tmp/mynewfile',
+    dest: '/tmp/mynewfile',
+    overwrite: false,
     callback: err => {
       if (err) console.error(err)
       else console.log('myfile success!')
     }
   },
 
-  '/tmp/mysecondfile': {
-    target: '/tmp/mysecondnewfile'
-  }
+  '/tmp/mysecondfile': '/tmp/mysecondnewfile'
 }
 
-fs.multi('copy', fileOptions, () => console.log('multiple processes completed'))
+fs.multi('copy', files, () => console.log('multiple processes completed'))
 ```
 
 Methods
--- a/README.md
+++ b/README.md
@@ -102,6 +102,30 @@ async function copyFiles () {
 copyFiles()
 ```
 
+Working with Multiple Files
+----------
+
+Examples:
+
+```js
+const fs = require('fs-extra')
+
+let fileOptions = {
+  '/tmp/myfile': {
+    target: '/tmp/mynewfile',
+    callback: err => {
+      if (err) console.error(err)
+      else console.log('myfile success!')
+    }
+  },
+
+  '/tmp/mysecondfile': {
+    target: '/tmp/mysecondnewfile'
+  }
+}
+
+fs.multi('copy', fileOptions, () => console.log('multiple processes completed'))
+```
 
 Methods
 -------
@@ -117,6 +141,7 @@ Methods
 - [mkdirp](docs/ensureDir.md)
 - [mkdirs](docs/ensureDir.md)
 - [move](docs/move.md)
+- [multi](docs/multi.md)
 - [outputFile](docs/outputFile.md)
 - [outputJson](docs/outputJson.md)
 - [pathExists](docs/pathExists.md)
@@ -135,6 +160,7 @@ Methods
 - [mkdirpSync](docs/ensureDir-sync.md)
 - [mkdirsSync](docs/ensureDir-sync.md)
 - [moveSync](docs/move-sync.md)
+- [multiSync](docs/multiSync)
 - [outputFileSync](docs/outputFile-sync.md)
 - [outputJsonSync](docs/outputJson-sync.md)
 - [pathExistsSync](docs/pathExists-sync.md)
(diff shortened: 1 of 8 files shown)


RyanZim:
I'm not seeing how these methods are useful vs. just making multiple copy/move calls. Also, if you're doing async operations, you could get a large performance boost by performing multiple calls in parallel, but that's only possible in some circumstances. As `multi()` currently stands, it's a performance footgun; and allowing the user to properly specify parallelism will almost certainly make the API too complex.

EsTharian:
> I'm not seeing how these methods are useful vs. just making multiple copy/move calls. Also, if you're doing async operations, you could get a large performance boost by performing multiple calls in parallel, but that's only possible in some circumstances. As `multi()` currently stands, it's a performance footgun; and allowing the user to properly specify parallelism will almost certainly make the API too complex.

I don't want to repeat myself while coding as:
```javascript
fs.copy('src/public', 'build')
fs.copy('src/view/', 'build/public', {overwrite: false}, err => console.error(err))
fs.copy('src/server.js', 'build/server.js')
fs.copy('src/routes', 'build/routes')
fs.copy('src/classes', 'build/classes')
fs.copy('src/bin', 'build/bin')
```

I think, DRY coding is more useful, especially working with hundreds of files:

```javascript
fs.multi('copy', {
	'src/public': 'build',
	'src/view/': {
            dest: 'build/public',
            opts: {
                overwrite: false
            },
            callback: err => console.error(err)
        },
	'src/server.js': 'build/server.js',
	'src/routes': 'build/routes',
	'src/classes': 'build/classes',
	'src/bin': 'build/bin'
})
```

I didn't test for perfomance but it should be a gain that is not calling the same function repeatedly.

RyanZim:
You're right that repeating yourself isn't optimal. However, there are very simple ways to get the same effect without needing a special method on `fs-extra`.

---

It's clear to me you're not understanding performance best practices for async operations. In your first example, you're not awaiting anything, and as a result there's no error handling.

Your current implementation of `fs.multi` has an `await` in a `for` loop. What  this means is that it will copy the first file, wait until that finishes, then begin copying the next file, wait for that, and so on. You would get the same effect by doing:

```js
await fs.copy('src/public', 'build')
await fs.copy('src/view/', 'build/public', {overwrite: false}, err => console.error(err))
...
```

This is _not_ the way to do things. What you want to do is start all the copy operations in parallel, and then wait for them all to finish:

```js
await Promise.all([
  fs.copy('src/public', 'build'),
  fs.copy('src/view/', 'build/public', { overwrite: false }),
  fs.copy('src/server.js', 'build/server.js'),
  fs.copy('src/routes', 'build/routes'),
  fs.copy('src/classes', 'build/classes'),
  fs.copy('src/bin', 'build/bin')
])
```

However, this still isn't DRY. You could actually use `fs.copy`'s `filter` option here for most of your copy operations:

```js
await Promise.all([
  fs.copy('src/public', 'build'),
  fs.copy('src/view/', 'build/public', { overwrite: false }),
  fs.copy('src', 'build', {
    filter: (src) => {
      return ['server.js', 'routes', 'classes', 'bin'].some((name) =>
        src.startsWith(path.join('src', name))
      )
    }
  })
])

```

EsTharian:
> You're right that repeating yourself isn't optimal. However, there are very simple ways to get the same effect without needing a special method on `fs-extra`.
> 
> It's clear to me you're not understanding performance best practices for async operations. In your first example, you're not awaiting anything, and as a result there's no error handling.
> 
> Your current implementation of `fs.multi` has an `await` in a `for` loop. What this means is that it will copy the first file, wait until that finishes, then begin copying the next file, wait for that, and so on. You would get the same effect by doing:
> 
> ```js
> await fs.copy('src/public', 'build')
> await fs.copy('src/view/', 'build/public', {overwrite: false}, err => console.error(err))
> ...
> ```
> 
> This is _not_ the way to do things. What you want to do is start all the copy operations in parallel, and then wait for them all to finish:
> 
> ```js
> await Promise.all([
>   fs.copy('src/public', 'build'),
>   fs.copy('src/view/', 'build/public', { overwrite: false }),
>   fs.copy('src/server.js', 'build/server.js'),
>   fs.copy('src/routes', 'build/routes'),
>   fs.copy('src/classes', 'build/classes'),
>   fs.copy('src/bin', 'build/bin')
> ])
> ```
> 
> However, this still isn't DRY. You could actually use `fs.copy`'s `filter` option here for most of your copy operations:
> 
> ```js
> await Promise.all([
>   fs.copy('src/public', 'build'),
>   fs.copy('src/view/', 'build/public', { overwrite: false }),
>   fs.copy('src', 'build', {
>     filter: (src) => {
>       return ['server.js', 'routes', 'classes', 'bin'].some((name) =>
>         src.startsWith(path.join('src', name))
>       )
>     }
>   })
> ])
> ```

OK. Shame... You are right about the loop.

I gave examples for creating a scenario. My actual point was DRY. Filter is the good pick for this scenario. But, if I have many different src and dest, I need loops or repeatings.

RyanZim:
Simple loops aren't hard to implement yourself; and I'm not convinced this is something that is frequently needed. Sorry.

PR #733, opened 2019-11-15
Make filesystem injectable for use with virtual FSes

### LINKED ISSUE #274: Use with memory-fs?

zetlen:
🚨DO NOT MERGE until https://github.com/jprichardson/node-jsonfile/pull/125 is merged. To make this all work, I had to propagate the same pattern to the `jsonfile` library, so this PR won't work unless you check out that PR and `npm link` it (instructions below)

Fixes #274.

We wanted to use the well-beloved `fs-extra` API in build systems that use a virtual FS under the hood for streaming and pluggability. This might include Webpack, Gulp, or other cool things we might use in the future. The nice thing about using `fs-extra` is that it already has a well-thought-out, well-maintained, and useful API.

However, you can't attach it to a different filesystem object without doing some dangerous monkeypatching.

This PR makes only small logic changes, though they look like large code changes. Instead of the individual module files creating methods based on the `graceful-fs` module directly, they return factory functions which receive an `fs` object to use.

#### Example of the change
(Highly simplified version of `ensureFile` for illustration purposes)

**Before:**
```js
const fs = require('graceful-fs')

module.exports = function createFile(file, callback) {
  fs.writeFile(file, '', callback)
}
```

**After:**
```js
module.exports = function createFileFactory(fs) {
  return function createFile(file, callback) {
    fs.writeFile(file, '', callback)
  }
}
```

This pattern is spread across the codebase so that the `fs` object can be injected from the top down.

This may cause concerns about performance of the constructor. This PR includes memoization of the method factories, based on the reference identity of the `fs` object being pushed in. This means that while you can create many `FsExtra` objects using the exact same filesystem, the methods themselves will be declared and constructed only once.

All tests pass on all supported OSes. The one area of concern is `utimes`, which uses `os.tmpdir()` to briefly test file timestamps. In a virtual filesystem, `os.tmpdir()` might not exist, so the test methods simply return false here, which is probably safest--virtual filesystems are not guaranteed to do timestamping right.

#### Verification

1. Pull and check out this branch and run `npm install`
2. In a sibling directory, clone the `node-jsonfile` dependency:

   ```sh
   git clone https://github.com/zetlen/node-jsonfile.git && \
   cd node-jsonfile && \
   git checkout zetlen/fs-injection && \
   npm install && \
   npm link
   ```

3. Complete the link in your `node-fs-extra` directory:

   ```sh
   npm link jsonfile
   ```

4. Run test suites and user acceptance experiments

Remember that until https://github.com/jprichardson/node-jsonfile/pull/125 is merged and released, you'll need to do that procedure after every reinstall of the node modules!


RyanZim:
I reviewed the PR for jsonfile. I'm fine with doing this for jsonfile, but the more I think about it, the less I like the idea of doing custom fs support here in fs-extra. Supporting three operating systems, with their various filesystems, and all their idiosyncrasies is bad enough. Then we have to support electron's `fs` module, which doesn't always behave the same as Node. If we add custom fs support, we open the floodgates to working around infinite inconsistencies in all kinds of filesystems. In short, it's asking for a nightmare. I'm gonna close this PR, and the issue here about custom fs support. Sorry for the wasted effort here @zetlen; I wish I had made this decision before you went to the work of doing a PR.

zetlen:
Don't worry about it, @RyanZim . If I was allergic to wasted effort, I would never open pull requests.

If you're still considering it, I think you can mitigate the support burden with an explicit list of supported alternate file systems and their versions. I think `memfs` and its cousins , as well as `memory-fs` to support legacy Webpack, would be plenty. And those are the ones I already tested!

Most people want to use one of the above file systems, and I think it's okay to state that:
- other alternate file systems are unsupported
- Issues about other file systems must come with a PR
- PRs to support other file systems must come with full regressions on all other supported file systems and use cases

Having said that, my work doesn't have a hard dependency on this, and I understand your reasoning.

PR #736, opened 2019-12-13
fix flaky test in remove.test.js

nacd:
Hi,

The async operations of fse.remove, fse.pathExists, and setInterval may interleave (race) causing the callback of fse.pathExists be called twice with variable itDoes=false. In this case, error `Error: done() called multiple times` is thrown, making the test flaky.


RyanZim:
Thanks!

PR #747, opened 2020-01-31
Update fs promise shims to support latest Node functions

RyanZim:
Added:

- `fs.writev`
- `fs.opendir`

### DIFF
--- a/docs/fs-read-write-writev.md
+++ b/docs/fs-read-write-writev.md
@@ -43,6 +43,6 @@ async function example () {
 ```js
 // With async/await:
 async function example () {
-  const { bytesWritten, buffers } = await fs.write(fd, buffers, position)
+  const { bytesWritten, buffers } = await fs.writev(fd, buffers, position)
 }
 ```
--- a/docs/fs-read-write-writev.md
+++ b/docs/fs-read-write-writev.md
@@ -1,6 +1,6 @@
 # About `fs.read()` & `fs.write()`
 
-[`fs.read()`](https://nodejs.org/api/fs.html#fs_fs_read_fd_buffer_offset_length_position_callback) & [`fs.write()`](https://nodejs.org/api/fs.html#fs_fs_write_fd_buffer_offset_length_position_callback) are different from other `fs` methods in that their callbacks are called with 3 arguments instead of the usual 2 arguments.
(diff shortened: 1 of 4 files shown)


docs/fs-read-write-writev.md
```diff
@@ -37,3 +37,12 @@ async function example () {
   const { bytesWritten, buffer } = await fs.write(fd, Buffer.alloc(length), offset, length, position)
 }
 ```
+
+## `fs.writev()`
+
+```js
+// With async/await:
+async function example () {
+  const { bytesWritten, buffers } = await fs.write(fd, buffers, position)
```
manidlou: This should be `fs.writev()`.

lib/fs/__tests__/multi-param.test.js
```diff
@@ -152,3 +154,59 @@ describe('fs.write()', () => {
...
+  afterEach(() => {
+    return fs.close(TEST_FD)
+      .then(() => fs.remove(TEST_FILE))
+  })
+
+  describe('with promises', () => {
+    it('returns an object', () => {
+      return fs.writev(TEST_FD, TEST_DATA, 0)
+        .then(({ bytesWritten, buffers }) => {
+          assert.strictEqual(bytesWritten, SIZE, 'bytesWritten is correct')
```
manidlou: I believe the message should be the opposite, something like `bytesWritten is incorrect` since that is used as the message for `AssertionError` if the values are not strictly equal. It is the same for other assertion messages down below.

https://nodejs.org/api/assert.html#assert_assert_strictequal_actual_expected_message

lib/fs/index.js
```diff
@@ -103,6 +105,25 @@ exports.write = function (fd, buffer, ...args) {
...
+  // Function signature is
+  // s.writev(fd, buffers[, position], callback)
+  // We need to handle the optional arg, so we use ...args
+  exports.writev = function (fd, buffers, ...args) {
+    if (typeof args[args.length - 1] === 'function') {
+      return fs.writev(fd, buffers, ...args)
+    }
+
+    return new Promise((resolve, reject) => {
+      fs.writev(fd, buffers, ...args, function (err, bytesWritten, buffers) {
```
manidlou: You don't need to change it now, but I realized we use `function` and `=>` very inconsistently in this file!

lib/fs/__tests__/multi-param.test.js
RyanZim: Yeah, perhaps; I don't think it really matters, the main point is to clearly identify which assertion is throwing if you have an error. In any case, these are simply copied from tests earlier in the file, so if we want to fix this, it should be in a separate PR that fixes the entire file.

PR #757, opened 2020-02-06
Cleanup lib/util/ to remove unused code

RyanZim:
Move function only used in tests to test files to avoid bundling it


RyanZim:
npm failed on one of the pr builds, but the push build is all good, so merging.

PR #774, opened 2020-03-23
catch => catch(error)

zxfrank:
Fixes error in catching error

`      } catch {
              ^
SyntaxError: Unexpected token {
`


zxfrank:
nevermind, updated node

PR #775, opened 2020-03-25
Adding parenthesis to prevent error

nathanhinish:
When I hit the catch I was getting a syntax error "Unexpected token {"


RyanZim:
You need to use Node v10 or higher to use fs-extra v9. Closing since this is intended.

nathanhinish:
Oh man, thanks for that. I was in an old Docker image that definitely needed updating.

PR #778, opened 2020-03-26
fix: add err params for catch function

janostudio:
fs-extra run in node 8.10.0 environment showing 'SyntaxError: Unexpected token {'.
Replace '} catch {' with '} catch (err) {' can solve the problem.


JPeer264:
Duplicate of #775 

We dropped the support for Node <10

Issue #759, opened 2020-02-11
Cannot move file to change case on case insensitive fs


wmertens:
When calling move() it checks if the destination exists, and if it does it refuses to call `rename`. This means you can't change the case on a case-insensitive file system.

Fs-extra should check that the destination has the same inode before refusing to move.

https://github.com/jprichardson/node-fs-extra/blob/40c3d68eb4fc4aa9a3c5624b4160610af296ec2e/lib/move/move.js#L41

manidlou:
@wmertens can you provide us with repro test case please?

wmertens:
The test itself is really simple: try to rename a file from lowercase to uppercase, it will fail when executed on a case-insensitive fs. The question is, can your test suite simulate that?

manidlou:
@wmertens I finally found some time to dig into your case! So here is what I found:

1. I appreciate if you would respect our issue template and respond to 3 simple questions that would help us resolve YOUR issue faster!
2. If you are using the latest version of `fs-extra`, we actually check the inodes of src and dest (as well as other properties) to determine if src and dest are the same before moving! However I just realized that implementation doesn't support cases like yours (when you want to just change the case) and I am working on that!
3. We actually have a [test file](https://github.com/jprichardson/node-fs-extra/blob/master/lib/move/__tests__/move-case-insensitive-paths.test.js) for cases like this, however again I just realized that those tests are broken for case-insensitive systems! Currently for case-insensitive systems those tests expect an error thrown however the corect behavior should allow users change the case for case-insensitive systems! and I am working on that too!

manidlou:
@wmertens I still don't understand how you get `dest already exists` error because with the latest version you get `Source and destination must not be the same` which is still not correct behavior for case-insensitive systems! That's why I asked you to give us an exact test case that would address your issue and you refused to do that too! So I am sorry from my perspective you don't have respect for this project and its maintainers' time!

PR #779, opened 2020-03-26
[breaking] copy*(): allow copying broken symlinks

### LINKED ISSUE #638: Broken symlinks should be copyable
- **Operating System:**
Mac OS High Sierra 10.13.6
- **Node.js version:**
8.11.3
- **`fs-extra` version:**
7.0.0, 8.0.0 (actually 8.0.0-6-gf3a2eed)

I understand the current implementation always resolves symlinks when initiating a copy, which has the side-effect of causing the copy to file if the symlink is broken. There are even tests enforcing that, even if `dereference` is set to `false`. It seems to me that this is undesirable behavior. In fact, if I copy a directory A that contains a symlink pointing to a file outside of it into a different location, it works, but the copy of the directory cannot itself be copied because it now hosts a broken symlink. The following code snippet illustrates this.

```
const fse = require('fs-extra')

// Setup:
// target
// sourceDir/
//   target-symlink -> ../target
fse.ensureFileSync('target')
fse.ensureDirSync('sourceDir')
fse.symlinkSync('../target', 'sourceDir/target-symlink')

// Act
// nestedDir/
//   sourceDirCopy/
//     target-symlink -> ../target (broken symlink)
fse.ensureDirSync('nestedDir')
fse.copySync('sourceDir', 'nestedDir/sourceDirCopy')

// And this is where it breaks due to the constraint that source symlink targets *must* exist
fse.copySync('nestedDir/sourceDirCopy', 'sourceDirCopy')
>> Error: ENOENT: no such file or directory, stat 'nestedDir/sourceDirCopy/target-symlink'
    at Object.fs.statSync (fs.js:948:11)
```

This behavior does not match that of `cp -R` which, while I understand is a completely different implementation, is oblivious of broken symlinks (at least in this context).

### LINKED ISSUE #761: Improve error message upon failed copy(file, dir)
- **Operating System: Win10**
- **Node.js version: v10.16.0**
- **`fs-extra` version: 8.1.0**

When you attempt to call `copy('./some-file', './some-dir')` you will get an error message that does not reflect what the docs are telling me not to do. The error message you get is about `unlink`, and I spent too much time tracking down what was really going on in a recent bit of code.

I now understand from issue #323 that it is *intentional* that a person should not try to call `copy()` with a file as the source and a directory as the destination, and I support the reasoning to have people be explicit in their intent. When I read the documentation for `copy()`, I see that this is explained clearly. But I wasn't the original author of the code and I wasn't the one reading the docs before writing the code. I was the debugger of code that came before me.

I propose that an error message be given that is consistent with the `copy()` documentation when the first arg is a file and the second arg is a directory. If that combination of arguments is documented to never be allowed, then the error message should reflect that.

### LINKED ISSUE #765: copySync throw error when directory has unreachable symlinks
- **Operating System:**
mac
- **Node.js version:**
v12.14.0
- **`fs-extra` version:**
8.1.0

error line:
https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy-sync/copy-sync.js#L121

function `copyDirItem` `stat.checkPathsSync` should be use fs.lstatSync when options:`dereference` is false

manidlou:
Fixes #765, fixes #638, fixes #761.


manidlou:
> This would be semver-major, right?

That's right @RyanZim.

manidlou:
@RyanZim should we wait for more approvals or this is ready to be merged?

RyanZim:
@manidlou I was holding off merging since it's semver-major, and I wasn't sure if I wanted to release a major for just this, or bundle in some other changes.

manidlou:
> I was holding off merging since it's semver-major, and I wasn't sure if I wanted to release a major for just this, or bundle in some other changes.

@RyanZim sounds good! So we can add this to the `10.0.0` milestone then.

manidlou:
> Do we have enough test cases where we use the actual default dereference: false? Seems like a lot of tests have changed to true. I'm not sure if these tests with the default false are actually needed

@JPeer264 most of those tests that are changed to `dereference: true` are edge cases that they needed to be changed because of the broken changes (new behavior). For the case `dereference: false` I added new test in the broken symlink test file. But generally, it doesn't hurt do have more tests for the case `dereference: false`!

manidlou:
@RyanZim I found a fix for #759 that will have a merge conflict with this later on. So I am gonna create `v10` branch and we will merge this to `v10` branch. That way I can move forward with the #759 fix.

Feel free to let me know if you have any objections!

manidlou:
@RyanZim @JPeer264 what is the deal with this branch `ci/travis-osximage`?! Do we still need to keep it?

JPeer264:
@manidlou thanks for mentioning that. I removed the branch, as it was no longer in use.

RyanZim:
@manidlou Just realized this was merged with a merge-commit into `v10`; do you mind if I recreate the branch with a squash-merge?

manidlou:
@RyanZim go for it.

RyanZim:
Major version branches are a pain in general, I've decided to just commit to it, and start landing semver-major stuff on master; landed in f21048b21d3806775560e4b112cddc54a1cb29d8.

Issue #246, opened 2016-05-05
fse.remove() && fse.removeSync() don't throw error on ENOENT file


kutyepov:
try {
    fse.removeSync('/directory/that/does/not/exist/index.js');
} catch (error) {
   //no exception will be thrown
}

Similarly,

fse.remove('/directory/that/does/not/exist/index.js', function(error) {
    //error is null
    console.error(error);
});

fse version 0.26.5 

Is it expected behavior?  Otherwise, I can help fixing it.

jprichardson:
> Is it expected behavior? Otherwise, I can help fixing it.

This is expected. Since you're trying to remove the file, why would you care if it doesn't already exist?

kutyepov:
Thank you for quick response. 

I would be just nice to throw an exception for debugging purposes e.g. if incorrect path was passed.

tndev:
I would also suggest to add an `ENOENT` error. 

There are situations where it is useful to get this information, e.g. when implementing a DAV base protocol or a RESTful interface, it is required to return an error if the client tires to delete a resource or collection that does not exits. As of that it is currently required to do a test if the file or directory exits before the `fse.remove` can be called. The race condition as described in the deprecated [fs.exists](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) would not be that problematic in the remove case, but it still would prefer not to check for the existence of the file in a separate call.

But main point why I would suggest to add the `ENOENT` is that `fse.remove` would behave same way as `fs.unlink`.

RyanZim:
> Since you're trying to remove the file, why would you care if it doesn't already exist?

Good point!

This would be a **huge** breaking change.

If this functionality is really needed (I'm not convinced it is), it should be a separate method. Feel free to open a feature request for that if you are so inclined.

CxRes:
Many years late to the discussion, could we add a loud ENOENT version under a flag ie `options: { AnnounceMissingObject: true }` . This won't be breaking in anyway afaict - all we need to do is change `cb(null)` to `options.AnnounceMissingObject ? cb(err) : cb(null)`. I could PR this if you like and would not mind suggestions on a better flag name.

RyanZim:
@CxRes What's the use-case? Seems like it's pretty much of a edge case, since you're the first person since 2016 to need this.

tndev:
[...]

PR #785, opened 2020-03-31
Remove: Option for Error on Missing File or Directory

CxRes:
As discussed in #246, this is a PR to allow for error message on ENOENT. I have provisionally only fixed `remove()` by adding a `options.shoutMissing` flag which is false by default.

I have made the changes defensively, since this is a library to fix the quirks of node fs and I am worried about breaking stuff. However, I have made more aggressive suggestions in the comments along side. Please examine and test it (I do not have access to all environments) and I'll make a proper PR according to your feedback.

### DIFF
--- a/docs/remove.md
+++ b/docs/remove.md
@@ -1,8 +1,13 @@
-# remove(path[, callback])
+# remove(path[, options][, callback])
 
-Removes a file or directory. The directory can have contents. If the path does not exist, silently does nothing. Like `rm -rf`.
+Removes a file or directory. The directory can have contents.
+
+If `options.shoutMissing` is not set and the path does not exist, silently does nothing. Like `rm -rf`.
 
 - `path` `<String>`
+- `options` `<Object>`
+  - `shoutMissing` `<boolean>`: Throw an error if path does not exist, default is `false`.
+  - `maxBusyTries` `<number>`: Maximum number of attempts at removal if file or directory is busy or locked, default is `3`.
 - `callback` `<Function>`
 
 ## Example:
--- a/lib/remove/__tests__/remove-dir.test.js
+++ b/lib/remove/__tests__/remove-dir.test.js
@@ -25,5 +25,13 @@ describe('remove / async / dir', () => {
         done()
       })
     })
+    it('should throw an error if shoutMissing is set', done => {
+      const someDir = path.join(TEST_DIR, 'some-dir/')
+      assert.strictEqual(fs.existsSync(someDir), false)
+      fse.remove(someDir, { shoutMissing: true }, err => {
+        assert.strictEqual(err.code, 'ENOENT')
+        done()
+      })
+    })
   })
 })
--- a/lib/remove/__tests__/remove.test.js
+++ b/lib/remove/__tests__/remove.test.js
@@ -120,5 +120,21 @@ describe('remove', () => {
         done()
       })
     })
+
+    it('should shout while not deleting glob matches when file doesn’t exist and shoutMissing is set', done => {
+      const nonexistentFile = path.join(TEST_DIR, 'file?')
+
+      const wrongFile = path.join(TEST_DIR, 'file1')
+      fs.writeFileSync(wrongFile, 'yo')
+
+      assert(!fs.existsSync(nonexistentFile))
+      assert(fs.existsSync(wrongFile))
+      fse.remove(nonexistentFile, { shoutMissing: true }, err => {
+        assert.strictEqual(err.code, 'ENOENT')
+        assert(!fs.existsSync(nonexistentFile))
+        assert(fs.existsSync(wrongFile))
+        done()
+      })
+    })
   })
 })
--- a/lib/remove/rimraf.js
+++ b/lib/remove/rimraf.js
@@ -22,6 +22,7 @@ function defaults (options) {
   })
 
   options.maxBusyTries = options.maxBusyTries || 3
+  options.shoutMissing = options.shoutMissing || false
 }
 
 function rimraf (p, options, cb) {
@@ -50,8 +51,12 @@ function rimraf (p, options, cb) {
         return setTimeout(() => rimraf_(p, options, CB), time)
       }
 
-      // already gone
-      if (er.code === 'ENOENT') er = null
+      if (er.code === 'ENOENT') {
+        // was never there
+        if (busyTries === 0) er = options.shoutMissing ? er : null
+        // already gone
+        else er = null
+      }
     }
 
     cb(er)
@@ -78,7 +83,7 @@ function rimraf_ (p, options, cb) {
   // so we have to lstat here and make sure it's not a dir.
   options.lstat(p, (er, st) => {
     if (er && er.code === 'ENOENT') {
-      return cb(null)
+      return options.shoutMissing ? cb(er) : cb(null) // just return cb(er); cb checks ENOENT
     }
 
     // Windows can EPERM on stat.  Life is suffering.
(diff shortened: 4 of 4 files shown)


CxRes:
@kutyepov Did you look at the aggressive changes (commented) as well?

kutyepov:
> 
> 
> @kutyepov Did you look at the aggressive changes (commented) as well?

mind elaborating on what you mean by 

> aggressive changes?

please?

CxRes:
@kutyepov As I wrote in the comments above, I was concerned that I may mess up the functionality so I made changes conservatively. However, next to the changes I made, I left comments with changes I think would make the code more efficient, hence aggressive! Could you please check if  that commented code might be preferable. Thanks in advance!!!!

kutyepov:
@CxRes it looks okay to me as it is. "Aggressive" changes will break backwards compatibility.

However, an opinion of maintainers will be more valuable and comprehensive than mine. I am just someone who ran into this issue years ago :)

CxRes:
@kutyepov Ok! Lets have one of the maintainers respond....

RyanZim:
Given that we'll probably be aiming for parity with Node core's new `recursive` option for `fs.rmdir` in the coming releases (see https://github.com/jprichardson/node-fs-extra/issues/806); I'd prefer not to add this additional option.

@manidlou @JPeer264 If you disagree, please reopen.

CxRes:
@ryanZim I am unable to understand your logic. The whole point of this library is to provide file-system functionality that node is unable to. I could use node's API if I just wanted parity! 

And it seems from the discussion in #246 that this is a popular request that many people have asked for. The reason offered for non-inclusion, quite incorrectly I might add, was that such a change would be breaking! This is **NOT** a breaking change!

RyanZim:
> The whole point of this library is to provide file-system functionality that node is unable to.

Until Node v12.10.0; there was no way to remove a directory with contents. So https://www.npmjs.com/package/rimraf was created to fill that gap. @jprichardson made this library because he got tired of installing rimraf and other similar utility modules on every project, so he bundled them all in this library. We now use a forked/vendored version of rimraf, but the codebase hasn't deviated much.

Now, finally Node core is beginning to add some of these common tasks to the core `fs` module. The `recursive` option for `fs.rmdir` is literally the code of rimraf being vendored in Node core. When that option is fully supported in all supported versions of Node (that'll be a few years from now), `remove()` won't have a reason to exist, since it will be providing the exact same functionality. We'll probably keep it as an alias for recursive `rmdir` for awhile, to allow people time to transition, but long-term, I don't see `remove()` being part of fs-extra. However, if we add this option, we'll be in an odd state where we'll be forced to keep a method that mirrors core by default, but has an extra option that changes behaviour. For that reason, I am opposed to adding this as an option to the current method.

---

In regards to adding this functionality in general (i.e. via a new method):

> this is a popular request that many people have asked for

Looking at https://github.com/jprichardson/node-fs-extra/issues/246, if I counted right, there's a grand total of 6 people in favour of it. For a package that's depended on by nearly 4 million other packages, I'm not convinced it qualifies as a _popular_ request. If we added every request that 6 developers asked for, we'd quickly end up with a massive, bloated, & complicated library. That said, I could be making the wrong call here. As I've stated before, if any of the maintainers disagree, I'd be happy to hear their point of view.

---

Thinking about this discussion, it does bring up an interesting point in my mind; it is a bit odd the way this works in Node core. Quoting Node docs:

> In recursive mode, errors are not reported if path does not exist, and operations are retried on failure.

There'd be a much stronger argument to say that Node core should not swallow `ENOENT` errors, from a standpoint of consistency. It's a bit odd that a single option makes a method **both** recursive _and_ ignore `ENOENT`. From a perspective of developer use, ignoring `ENOENT` is IMO the most common desired outcome, but from a consistency perspective, it's odd. *nix `rm` has separate options for this, `-r` & `-f` (`rm -rf` is just shorthand for `rm -r -f`). `-r` makes it recursive, `-f` makes it swallow `ENOENT`.

Might not be a dumb idea to open to open an issue on Node itself about this. If Node core changed functionality/added an option, my opinion on supporting this in `fs-extra` would change.

CxRes:
@RyanZim Thanks for a very detailed reply. I am more sympathetic to your thought process. But I still have to disagree. Let me see if I can explain my POV better: 

`fse.remove` in some sense is to node what `rm` would be to posix. Both combine the functionality of `unlink` and `rmdir`.  AFAICT the node API is essentially modeled after the posix API. Heck, they even link directly to linux documentation for most functions including `unlink` and `rmdir`. 

Now, I do not see how you can get rid of `fse.remove` given that the target, much like that for `rm`, can also be a file. The benefit of `remove` is that one need not know the state of the file system before requesting the operation. A recursive `fs.rmdir` still suffers from this problem and will force the check back to user land. If `rm` can inform of `ENOENT` if we want, imho `fse.remove` should too. I would be open to changing the form so that it is more `rm` like!

In any case, I would also welcome more input; we all want the best possible outcome!!!!  

Now to some of your comments:

> Thinking about this discussion, it does bring up an interesting point in my mind; it is a bit odd the way this works in Node core. 

I think the way node is approaching this is wrong. Having followed posix conventions all along, they seem to have suddenly abandoned it with these poorly thought out flags on `rmdir`. I shall open an issue with node sometime next week, though I am not sure how to broach this issue.... from the point of view of flag removal or adding an `rm` equivalent? Suggestions?

>  if I counted right, there's a grand total of 6 people in favour of it...

With all due respect, I cannot help but chuckle! You speak of gazillions of users, I submit to you that there have ever been only 500 odd issues filed and 300 PRs (and much fewer of that would be regarding feature requests). Very few users ever speak up and even fewer submit changes. Sure, popular may be overstating it, but in the context of those numbers, my "grand total" of 6 people over the span of a couple of years speaking up on the same matter might not just be incidental or accidental. Statistics can be used to trivialize anything!

> If we added every request that 6 developers asked for, we'd quickly end up with a massive, bloated, & complicated library.

Of course, I share your aversion to adding bloat but at the same time I believe that one should strive for feature completeness as well. It is difficult balance. However, in this case, a comparison with `rm` (which was made perhaps 25 years ago or more) indicates that this might not be a case of bloat.

RyanZim:
You're correct that so far, Node has mostly be modelled after POSIX. The logic being implemented for recursive `rmdir` is functionally the same as our `remove`. That is, it will work even if the target is a file. Now, it's clear that doesn't make much sense. `fs.rmdir(filename, { recursive: true })` does not convey intent properly at all.  Effectively, `recursive` option turns `rmdir` into `rm -f`. The more I think about this, the more I am convinced the current direction of Node's naming does not make any sense. Currently considering opening an issue with Node myself.

Re: your comments about statistics, fair points. Not 100% sure I agree, but haven't given the matter deep thought. Regardless, that's neither here nor there ATM, since I won't be moving forward here until I get clarity on what Node is doing/thinking.

CxRes:
Fair enough! Lets see what the folks at Node have to say....

CxRes:
@RyanZim I finally remembered to file that issue!!! Hope its OK that i took the liberty to link you in the discussion.

RyanZim:
Yes, thanks. Was meaning to do it myself, but didn't make it a priority. Thanks for going ahead; commented my thoughts there.

CxRes:
@RyanZim Node has now "fixed" recursive remove! The ball is in your court now...

RyanZim:
@CxRes Please file an issue so we can discuss potential APIs for doing this.

CxRes:
Please reopen #246!

RyanZim:
Fresh thread filed: https://github.com/jprichardson/node-fs-extra/issues/840

PR #794, opened 2020-05-05
Add `(e)` to catch blocks

loup-brun:
**I know node v8 is not in the list of engines.**
Nevertheless, this _very minor fix_ avoids failing on node v8, where everything else works perfectly.

Basically, it’s just adding the `(e)` on `catch` blocks which don’t specify an exception var.

```js
try {
  // ...
} catch (e) {}
```

This package is the only one failing on a _very lightweight project_, and my team finds it painful to download multiple node versions across projects (I don’t mind using NVM, but it’s still a lot of extra download)…

It would be very kind of you to integrate this small change.

### DIFF
--- a/lib/copy/__tests__/copy-permissions.test.js
+++ b/lib/copy/__tests__/copy-permissions.test.js
@@ -32,13 +32,13 @@ describe('copy', () => {
 
     try {
       gidWheel = process.getgid() // userid.gid('wheel')
-    } catch {
+    } catch (e) {
       gidWheel = process.getgid()
     }
 
     try {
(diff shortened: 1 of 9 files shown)


JPeer264:
Actually Node v8 has reached [EOL](https://nodejs.org/en/about/releases/). Since v9 we do not support Node v8 anymore, so you need to update to Node v10+ to use fs-extra v9+

@RyanZim feel free to reopen if there are any exceptions regarding this

RyanZim:
Yeah, Node v8 is EOL, so it should not be used, period. Also, the fact that fs-extra works on v8 otherwise is by chance, it may break further at any time.

PR #795, opened 2020-05-08
Improve docs

### LINKED ISSUE #753: Unix commands vs. fs-extra functions
Something that I keep seeing in users' comments is that users expect `fs-extra` functions to have the same behavior/implementation as unix commands `cp, mv, rm...`.

At least, I couldn't find anywhere in the docs that specifies we follow unix or posix or any other specs. 

I think it is a good idea to clarify this for our users. So my question is, what should we do here? Should we follow some specs or we should specify in the docs that we don't follow any specs?

manidlou:
Just a little improvement in docs! Resolves #753.

### DIFF
--- a/docs/ensureSymlink-sync.md
+++ b/docs/ensureSymlink-sync.md
@@ -6,8 +6,7 @@ Ensures that the symlink exists. If the directory structure does not exist, it i
 
 - `srcPath` `<String>`
 - `destPath` `<String>`
-- `type` `<String>`
-	- It is only available on Windows and ignored on other platforms. It can be set to `dir`, `file`, or `junction`.
+- `type` `<String>` It is only available on Windows and ignored on other platforms. It can be set to `dir`, `file`, or `junction`.
 
 ## Example:
 
--- a/docs/ensureSymlink-sync.md
+++ b/docs/ensureSymlink-sync.md
@@ -1,20 +1,21 @@
-# ensureSymlinkSync(srcpath, dstpath[, type])
+# ensureSymlinkSync(srcPath, destPath[, type])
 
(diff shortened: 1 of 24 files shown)


docs/ensureSymlink-sync.md
```diff
@@ -1,20 +1,21 @@
...
 Ensures that the symlink exists. If the directory structure does not exist, it is created.
 
 **Alias:** `createSymlinkSync()`
 
-- `srcpath` `<String>`
-- `dstpath` `<String>`
+- `srcPath` `<String>`
+- `destPath` `<String>`
 - `type` `<String>`
+	- It is only available on Windows and ignored on other platforms. It can be set to `dir`, `file`, or `junction`.
```
RyanZim, line 10:
```suggestion
- `type` `<String>` It is only available on Windows and ignored on other platforms. It can be set to `dir`, `file`, or `junction`.
```

docs/ensureSymlink.md
```diff
@@ -1,30 +1,32 @@
...
 Ensures that the symlink exists. If the directory structure does not exist, it is created.
 
 **Alias:** `createSymlink()`
 
-- `srcpath` `<String>`
-- `dstpath` `<String>`
+- `srcPath` `<String>`
+- `destPath` `<String>`
 - `type` `<String>`
+  - It is only available on Windows and ignored on other platforms. It can be set to `dir`, `file`, or `junction`.
```
RyanZim, line 10:
```suggestion
- `type` `<String>` It is only available on Windows and ignored on other platforms. It can be set to `dir`, `file`, or `junction`.
```

manidlou:
Sure @RyanZim! good call!

PR #801, opened 2020-05-16
move: support changing case in case-insensitive systems

### LINKED ISSUE #759: Cannot move file to change case on case insensitive fs

manidlou:
Resolves #759.

### DIFF
--- a/lib/move/move.js
+++ b/lib/move/move.js
@@ -30,13 +30,13 @@ function move (src, dest, opts, cb) {
 }
 
 function doRename (src, dest, overwrite, isChangingCase, cb) {
+  if (isChangingCase) return rename(src, dest, overwrite, cb)
   if (overwrite) {
     return remove(dest, err => {
       if (err) return cb(err)
       return rename(src, dest, overwrite, cb)
     })
   }
-  if (isChangingCase) return rename(src, dest, overwrite, cb)
   pathExists(dest, (err, destExists) => {
     if (err) return cb(err)
     if (destExists) return cb(new Error('dest already exists.'))
--- a/lib/move/move.js
+++ b/lib/move/move.js
@@ -18,24 +18,25 @@ function move (src, dest, opts, cb) {
 
   stat.checkPaths(src, dest, 'move', opts, (err, stats) => {
     if (err) return cb(err)
-    const { srcStat } = stats
+    const { srcStat, isChangingCase = false } = stats
     stat.checkParentPaths(src, srcStat, dest, 'move', err => {
       if (err) return cb(err)
       mkdirp(path.dirname(dest), err => {
         if (err) return cb(err)
-        return doRename(src, dest, overwrite, cb)
+        return doRename(src, dest, overwrite, isChangingCase, cb)
       })
     })
   })
 }
 
-function doRename (src, dest, overwrite, cb) {
+function doRename (src, dest, overwrite, isChangingCase, cb) {
   if (overwrite) {
     return remove(dest, err => {
       if (err) return cb(err)
       return rename(src, dest, overwrite, cb)
     })
   }
+  if (isChangingCase) return rename(src, dest, overwrite, cb)
   pathExists(dest, (err, destExists) => {
     if (err) return cb(err)
     if (destExists) return cb(new Error('dest already exists.'))
--- a/lib/util/stat.js
+++ b/lib/util/stat.js
@@ -11,8 +11,6 @@ const lstat = (file) => nodeSupportsBigInt ? fs.lstat(file, { bigint: true }) : 
 const statSync = (file) => nodeSupportsBigInt ? fs.statSync(file, { bigint: true }) : fs.statSync(file)
 const lstatSync = (file) => nodeSupportsBigInt ? fs.lstatSync(file, { bigint: true }) : fs.lstatSync(file)
 
-const isCaseInsensitiveSystem = process.platform === 'darwin' || process.platform === 'win32'
-
 function getStats (src, dest, opts) {
   const statFunc = opts.dereference ? stat : lstat
   return Promise.all([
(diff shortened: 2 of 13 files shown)


RyanZim:
@manidlou Did you verify this actually changes the case displayed in the file manager or on the command line, for Mac or Windows?

manidlou:
@RyanZim that's definitely required and I wanted to do that but I don't have neither of those systems. I appreciate if anyone could verify that on Mac and Windows! Feel free to let me know if the changes don't work and I will work on it then! I also might be able to have access to my friend's macbook tomorrow, so I'll let you know then.

manidlou:
I verified that on Mac and it successfully changed the case.

lib/move/move.js
```diff
@@ -18,24 +18,25 @@ function move (src, dest, opts, cb) {
...
+        return doRename(src, dest, overwrite, isChangingCase, cb)
       })
     })
   })
 }
 
-function doRename (src, dest, overwrite, cb) {
+function doRename (src, dest, overwrite, isChangingCase, cb) {
   if (overwrite) {
     return remove(dest, err => {
```
wmertens, line 34: If `overwrite: true` and the file just changes case, won't it then remove the file that it's trying to rename?

manidlou: Yeah I see your point! It is safer to check changing case before checking overwrite. However, I don't understand what would be the point of using `overwrite: true` when you just want to change the case!

Basically, the `overwrite: true` is useful when there is a possibility (or we know for sure) that dest already exists and we want to move src to dest regardless. In this case, we are dealing with 2 files and `overwrite: true` is needed in order to get the desired result. But, when we are just changing the case, we are basically dealing with 1 file, so there is no need to use the `overwrite` option!

wmertens: I can see this happening when a dev just wants to move a file to where some result says it should, and the result just changes the case.

I saw that you moved out, but maybe this is important enough to warrant a test?

manidlou: I verified that on Mac, but it doesn't hurt to have a test for it!

lib/util/stat.js
```diff
@@ -11,6 +11,8 @@ const lstat = (file) => nodeSupportsBigInt ? fs.lstat(file, { bigint: true }) :
 const statSync = (file) => nodeSupportsBigInt ? fs.statSync(file, { bigint: true }) : fs.statSync(file)
 const lstatSync = (file) => nodeSupportsBigInt ? fs.lstatSync(file, { bigint: true }) : fs.lstatSync(file)
 
+const isCaseInsensitiveSystem = process.platform === 'darwin' || process.platform === 'win32'
```
wmertens, line 14: Linux also has case-insensitive filesystems, and it can network mount CIFS which is also case-insensitive. So the only way to test this is to see if you can access a file on the filesystem you want to work with via another casing that doesn't exist yet.

lib/util/stat.js
```diff
@@ -42,6 +44,14 @@ function checkPaths (src, dest, funcName, opts, cb) {
 
     if (destStat) {
       if (areIdentical(srcStat, destStat)) {
+        const srcBaseName = path.basename(src)
+        const destBaseName = path.basename(dest)
+        if (funcName === 'move' &&
+          isCaseInsensitiveSystem &&
```
wmertens, line 50: perhaps better to simply not test for this. If the `stat()` result is the same, it's the same file.

lib/util/stat.js
manidlou: I agree that it doesn't include all case-insensitive file systems. Basically a better approach would be to actually probe the file system to find out about its case-sensitivity. It might have some performance drawback but ultimately that would be a better approach.

wmertens: But wouldn't you have to do that for every directory you access?

lib/util/stat.js
wmertens: ... And in that case, just call rename(). I can't think of a case where it would fail

lib/util/stat.js
manidlou: If directories are mounted on different file systems, yes! Basically, finding case-sensitivity of a file system is a complex thing! If we want to be 100% safe, we need to check where the directories are mounted, and also if there is any subsystem that has its own case comparison table.

wmertens: So wouldn't it be easier to just rename the file and pass any errors to the user instead?

RyanZim:
@manidlou ping!

lib/util/stat.js
manidlou: That makes sense! I am ok with not checking for case sensitivity of the system as this is not comprehensive enough!

manidlou:
Sorry for the long pause! Been extremely busy! I will have a look at this again.

RyanZim (commented):
LGTM; except test descriptions need updated, since there's nothing OS-specific anymore (shouldn't say "based on the OS").

manidlou:
Since this will merge into v10, and I am a bit worried if we wait for long time, we might end up having some merge conflicts! So I merged current master into v10 and rebased this one so that we will have a clean merge!

RyanZim:
@manidlou Honestly, I suspect I'll end up manually cherry-picking everything related to v10, since it's taken so long. Just leave this PR as-is for now.

RyanZim:
Landed in https://github.com/jprichardson/node-fs-extra/commit/c8bd3830fc19ab1367dbe2169e0bbde052475a89.

Issue #804, opened 2020-06-03
When using jest there seems to be a problem when deleting a file or a folder on Windows


federicobozzini:
- **Operating System:**Windows 10
- **Node.js version:**10.16
- **`fs-extra` version:**9.0.0

When using jest it seems there might be a problem when deleting a file or a folder on Windows . The error seems to stem from the rimraf-like code. See https://github.com/isaacs/rimraf/issues/208.

RyanZim:
Fixed in https://github.com/jprichardson/node-fs-extra/pull/805.

PR #805, opened 2020-06-03
fix: do not check errors with `instanceof`.

federicobozzini:
This fixes cross-VM type issues in Jest #804 .

See https://github.com/isaacs/rimraf/commit/aa50e0298f66b3375045e8d9d86992d696a56b09


RyanZim:
Published in 9.0.1 :tada:

Issue #807, opened 2020-06-08
Add param to catch blocks


Qunapi:
When I test fs-extra in my node typescript app, it can't compile code due to a syntax error
According to the [standart](http://es5.github.io/#x12.14) , after catch there always must be error variable 
example: 
https://github.com/jprichardson/node-fs-extra/blob/7b12b058e27df560ba777756f38f977662c23750/lib/mkdirs/make-dir.js#L85 
```
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/qunapi/gitdb/node_modules/fs-extra/lib/mkdirs/make-dir.js:86
          } catch {
                  ^

    SyntaxError: Unexpected token {

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1258:14)
      at Object.<anonymous> (node_modules/fs-extra/lib/mkdirs/index.js:3:44)
```

RyanZim:
See https://github.com/jprichardson/node-fs-extra/pull/794

PR #808, opened 2020-06-08
fix: add param to catch blocks

Qunapi:
Fixes [#807](https://github.com/jprichardson/node-fs-extra/issues/807)


RyanZim:
Dupe of https://github.com/jprichardson/node-fs-extra/pull/794.

PR #812, opened 2020-06-16
Error when unknown file type is encountered

ibash:
fs.copy was silently exiting when a socket was encountered in my
directory. This prevents silent failure so the user can deal with using
fs.copy appropriately.

### DIFF
--- a/lib/copy-sync/copy-sync.js
+++ b/lib/copy-sync/copy-sync.js
@@ -47,6 +47,7 @@ function getStats (destStat, src, dest, opts) {
            srcStat.isCharacterDevice() ||
            srcStat.isBlockDevice()) return onFile(srcStat, destStat, src, dest, opts)
   else if (srcStat.isSymbolicLink()) return onLink(destStat, src, dest, opts)
+  else { cb(new Error(`Unknown file: ${src}`)) }
 }
 
 function onFile (srcStat, destStat, src, dest, opts) {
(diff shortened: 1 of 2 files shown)


RyanZim:
Is this technically `semver-major`, since we're turning a buggy silent exit into an error?

ibash:
Updated `copySync`

> Is this technically semver-major, since we're turning a buggy silent exit into an error?

I think it could be argued both ways. On one hand semver is about having a public api. In this instance this code broke the conventional contract that callbacks are always called, so it's correcting undefined behavior.

On the other hand it really depends on what users of the library are expecting. I got here via stepping through `electron-packager` code until I found where execution stopped. `electron-packager` definitely wasn't expecting callbacks to not be called (there's actually a number of bug reports about it exiting halfway because of things like this).

The other major users are https://github.com/webpack-contrib/eslint-loader https://github.com/GoogleChrome/workbox and https://github.com/danethurber/webpack-manifest-plugin.

So maybe @ricardogobbosouza @jeffposnick and @DanielRuf?

DanielRuf:
Not sure why I was mentioned. Can you clarify this @ibash?

DanielRuf:
If it was because of https://github.com/danethurber/webpack-manifest-plugin/commit/b55ac5d7748dde89fd0225e13263c7b42b1dca87, I am just a normal contributor and contributed a PR for the CI setup. And that's all ;-)

So I think this was by accident.

DanielRuf:
I think you mentioned me by accident. I was only a normal contributor for a small CI related PR, nothing more: https://github.com/danethurber/webpack-manifest-plugin/commit/b55ac5d7748dde89fd0225e13263c7b42b1dca87

DanielRuf:
Besides this, turning a silent error (warning or notice in other frameworks and languages) to an actual (fatal) error is a breaking change as an error breaks the program flow.

ibash:
Ah my bad, I just mentioned the last author on master :)

PR #814, opened 2020-06-16
Fixing a syntax error of a try/catch block

vandervidi:
added missing error object to catch


RyanZim:
Duplicate of https://github.com/jprichardson/node-fs-extra/pull/794.

PR #826, opened 2020-07-29
BREAKING: Ensure existing dest correct type/target in ensureLink/Symlink

### LINKED ISSUE #786: ensureSymlink silently exits if the destination exists and is not a symlink
Looking into the code, I found that ensure* methods only verify if the destination path exist and not if the path is of the right * type. IMHO an error should be thrown if path is of the wrong type. Perhaps `fs.stat` instead of `fs.access` should suffice...

### LINKED ISSUE #870: `ensureLink` does not ensure the destination is hard linked to the source
- **Operating System:** macOS 10.15.7
- **Node.js version:** v12.20.1
- **`fs-extra` version:** 5.0.0 (checked source code in `master` and the issue is still there)

`ensureLink` shortcuts if the destination path exists, without verifying if the destination file's inode is the same as the source file. This is misleading because it is not really an "ensurance".

Relevant code: https://github.com/jprichardson/node-fs-extra/blob/6bffcd81881ae474d3d1765be7dd389b5edfd0e0/lib/ensure/link.js#L19

RyanZim:
Fixes #786 
Fixes #870

CC @cxres

### DIFF
--- a/lib/ensure/link.js
+++ b/lib/ensure/link.js
@@ -5,6 +5,7 @@ const path = require('path')
 const fs = require('graceful-fs')
 const mkdir = require('../mkdirs')
 const pathExists = require('../path-exists').pathExists
+const { areIdentical } = require('../util/stat')
 
 function createLink (srcpath, dstpath, callback) {
   function makeLink (srcpath, dstpath) {
@@ -14,14 +15,13 @@ function createLink (srcpath, dstpath, callback) {
     })
   }
 
-  pathExists(dstpath, (err, destinationExists) => {
-    if (err) return callback(err)
-    if (destinationExists) return callback(null)
-    fs.lstat(srcpath, (err) => {
+  fs.lstat(dstpath, (_, dstStat) => {
+    fs.lstat(srcpath, (err, srcStat) => {
       if (err) {
         err.message = err.message.replace('lstat', 'ensureLink')
         return callback(err)
       }
+      if (dstStat && areIdentical(srcStat, dstStat)) return callback(null)
 
       const dir = path.dirname(dstpath)
       pathExists(dir, (err, dirExists) => {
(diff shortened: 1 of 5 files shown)


lib/ensure/link.js
```diff
@@ -14,9 +14,9 @@ function createLink (srcpath, dstpath, callback) {
     })
   }
 
-  pathExists(dstpath, (err, destinationExists) => {
-    if (err) return callback(err)
-    if (destinationExists) return callback(null)
+  fs.lstat(dstpath, (err, stats) => {
+    if (!err && stats.isSymbolicLink()) return callback(null)
```
manidlou, line 18: This doesn't work for links (hard links) since `stats.isSymbolicLink()` would return `false` for hard links as symlinks and hard links are different types. The thing is, there is no easy way to check if a file is a hard link! One way to do that would be by checking the entire file system :exploding_head:  to see if there is any file that has the same inode as the hard link since hard link and the original file have the same inodes on the file system.

I am personally not sure what would be the best approach here.

1. We could check if dest exists and if is symlink, then throw error.
2. Or we allow the operation proceeds which means if dest exists, a `file already exists` error will be thrown.

Let me know what you guys think!

RyanZim: Ah, yeah, unit tests should be modified to cover this case, I didn't think this through.

Actually, I'm specifically wondering if we shouldn't be checking that it's a hard link between given src and dest, since we're trying to ensure it. Same actually goes for symlinks.

RyanZim:
@manidlou Finally got around to updating this; now checks both the type of the dest, and also ensures the dest matches the source as requested. If the dest exists, but is the wrong type, or points to the wrong place, allows it to error out.

manidlou (approved):
LGTM

PR #837, opened 2020-10-09
Update make-dir.js

jacobbe:
catch without variable block


RyanZim:
Duplicate of https://github.com/jprichardson/node-fs-extra/pull/794.

PR #839, opened 2020-10-14
Fix syntax errors that will cause SyntaxError: Unexpected token

jcplus:
This project is depended by other projects and throws **SyntaxError: Unexpected token** errors while initialisation. For example: [raxjs/rax-scripts](https://github.com/raxjs/rax-scripts). I've added all the misisng `(error)` bits. Examples below:

```
/usr/local/lib/node_modules/rax-cli/node_modules/fs-extra/lib/mkdirs/make-dir.js:85
      } catch {
              ^
SyntaxError: Unexpected token {
```

```
/usr/local/lib/node_modules/rax-cli/node_modules/fs-extra/lib/mkdirs/make-dir.js:134
      } catch {
              ^
SyntaxError: Unexpected token {
```

### DIFF
--- a/lib/copy/__tests__/copy-permissions.test.js
+++ b/lib/copy/__tests__/copy-permissions.test.js
@@ -32,13 +32,13 @@ describe('copy', () => {
 
     try {
       gidWheel = process.getgid() // userid.gid('wheel')
-    } catch {
+    } catch (error) {
       gidWheel = process.getgid()
     }
 
     try {
       gidStaff = process.getgid() // userid.gid('staff')
-    } catch {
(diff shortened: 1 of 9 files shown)


JPeer264:
This is intended, as we [only support Node v10+](https://github.com/jprichardson/node-fs-extra/blob/master/package.json#L6) with `fs-extra` v9+.  

It seems that the library you mentioned is also only testing (therefore I assume only supporting) Node v10(+), as you can see here: https://github.com/raxjs/rax-scripts/blob/master/.github/workflows/ci.yml#L12

PR #848, opened 2020-11-19
Fixing catch exception syntax

paulodrade:


JPeer264:
Thanks for contributing, but we decided to leave them as we support only Node v10+ in fs-extra v9+

See #794

PR #852, opened 2020-12-04
copy: support mode for copyFile

rickycao-qy:
This PR is to support reflink when copying files.

### DIFF
--- a/docs/copy-sync.md
+++ b/docs/copy-sync.md
@@ -9,6 +9,10 @@ Copy a file or directory. The directory can have contents.
   - `errorOnExist` `<boolean>`: when `overwrite` is `false` and the destination exists, throw an error. Default is `false`.
   - `dereference` `<boolean>`: dereference symlinks, default is `false`.
   - `preserveTimestamps` `<boolean>`: When true, will set last modification and access times to the ones of the original source files. When false, timestamp behavior is OS-dependent. Default is `false`.
+  - `mode` `<integer>`: modifiers for copy operation, default is `0`.
+    - `fs.constants.COPYFILE_EXCL` - The copy operation will fail if dest already exists.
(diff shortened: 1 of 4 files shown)


RyanZim:
`mode` is only supported by Node 14+; and some of the options here overlap with some of our existing options. Long-term, I think moving to an API closer to `copyFile` makes sense, but I'm not sure if now is the right time. @manidlou thoughts?

rickycao-qy:
> `mode` is only supported by Node 14+; and some of the options here overlap with some of our existing options. Long-term, I think moving to an API closer to `copyFile` makes sense, but I'm not sure if now is the right time. @manidlou thoughts?

Node.js with older version has the [optional parameter flag](https://nodejs.org/docs/latest-v10.x/api/fs.html#fs_fs_copyfile_src_dest_flags_callback) which specifies modifiers for the copy operation. Can we just make this uniform in copy API

rickycao-qy:
@manidlou @RyanZim Could you have a look, please

manidlou (commented):
Sorry for late reply! I also agree that it is generally a good idea to be as compatible as possible with node core!

So, this option only applies to copying files, however `fs-extra` tries to abstract away the difficulties of differences between files and dirs for the end user.

IMHO, @rickycao-qy  we need to come up with a better name than `mode` for this option since it is not descriptive enough to let our users know this option is only used for copying files.

rickycao-qy:
> Sorry for late reply! I also agree that it is generally a good idea to be as compatible as possible with node core!
> 
> So, this option only applies to copying files, however `fs-extra` tries to abstract away the difficulties of differences between files and dirs for the end user.
> 
> IMHO, @rickycao-qy we need to come up with a better name than `mode` for this option since it is not descriptive enough to let our users know this option is only used for copying files.

Hey, thanks for the reply. Personally I think this option is useful to copy dirs as we just apply this option to recursively copy files in dirs when applying to dirs. Or we can use a mofe descriptive option such as `copyFileMode`?

manidlou:
I am ok with `copyFileMode`.

RyanZim:
Closing as this is stale.

PR #860, opened 2021-01-18
Add promise support for fs.rm()

### LINKED ISSUE #841: Promisify fs.rm
Related to https://github.com/jprichardson/node-fs-extra/issues/806, but making it a separate issue to be explicit.

We need to add `fs.rm` to our list of fs methods to add promise support to.

RyanZim:
Fixes #841

### DIFF
--- a/lib/fs/__tests__/rm.test.js
+++ b/lib/fs/__tests__/rm.test.js
@@ -1,0 +1,30 @@
+'use strict'
+
+const fse = require('../..')
+const os = require('os')
+const path = require('path')
+const assert = require('assert')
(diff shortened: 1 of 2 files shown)


ChALkeR (commented):
utACK.

When I attempted to test, other local tests fail on master, investigating.

Side note: not sure why this code needs `universalify` dep at all, since this already has Node.js 10 in engines, and Node.js >=8 has `util.promisify` support. Will check separately.

ChALkeR:
Q: why not just export `fs.promises.rm`?

There is no version of Node.js where `fs.rm` is available and `fs.promises.rm` isn't available.

PR #866, opened 2021-02-03
Fixes functions outputJSON and pathExists execute callbacks even if there are errors/missing args

### LINKED ISSUE #865: [bug] functions outputJSON and pathExists execute callbacks even if there are errors/missing args (not matching fs behaviour)
Hi all! I'm working on an automated test generator and have been testing it by applying regression testing to a set of popular npm libraries, including yours! In [this commit](https://github.com/jprichardson/node-fs-extra/commit/a149f822c387958caba7333ad2830615ab8b078f) I found a difference in the behaviour of `outputJSON` when called with incorrect arguments. 

### The issue
Functions that have been converted with `universalify.fromPromise` execute the callback argument even when the other arguments are missing or invalid. It looks like this results in a bug, because the comparable `fs` functions do error checking of these arguments.

#### To replicate the issue:
```
let fse = require('fs-extra');
function callback() { 
  console.log("Callback executing!");
  console.log(arguments);
}
fse.outputJSON("newfile.json", "[ ]", callback); // correct call
fse.outputJSON( callback); // incorrect call
```
Before the change in [this commit](https://github.com/jprichardson/node-fs-extra/commit/a149f822c387958caba7333ad2830615ab8b078f),  the output of running the above code is:
```
// correct call output: 
> Callback executing!
> [Arguments] { '0': null, '1': undefined }
// incorrect call output:
> Uncaught:
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received function callback
    at validateString (internal/validators.js:124:11)
    at Object.dirname (path.js:1128:5)
...
```
Then, after the update, the output of running the above code is:
```
// correct call output:
> Callback executing!
> [Arguments] { '0': null, '1': undefined }
// incorrect call output:
> Callback executing!
> [Arguments] {
  '0': TypeError: Cannot read property 'replace' of undefined
      at stringify ...
```
So, the callback is now executed even when the arguments are incorrect, and have the function error passed in as an argument to the callback.

### What is causing this behaviour
[This change](https://github.com/jprichardson/node-fs-extra/commit/a149f822c387958caba7333ad2830615ab8b078f#diff-aaca9f40ebdd867e7cf75dcf5ed1d65a857fc9cfe1238bb52f7461028a9af0afR3) from using  the [universalify](https://www.npmjs.com/package/universalify) library’s `fromPromise` function instead of `fromCallback` is causing the difference.

Looking at [the implementation of `fromPromise`](https://github.com/RyanZim/universalify/blob/master/index.js#L18): this function takes a function `fn` as an argument, and returns a new function that
* calls `fn` as it would normally be called, [if the last argument is not a callback](https://github.com/RyanZim/universalify/blob/master/index.js#L21)
* [if the last argument _is_ a callback](https://github.com/RyanZim/universalify/blob/master/index.js#L22), then `fn` is called with all but its last argument (this is `fn.apply(this, args.slice(0, -1))`). As [stated in the documentation](https://github.com/RyanZim/universalify#universalifyfrompromisefn), the `fromPromise` function assumes that `fn` returns a promise.  Then, it chains a `.then` to the return of the call, passing the function `r => cb(null, r)` as the resolve of this `.then` promise, and `cb` (the callback argument itself) as the reject function.

In our case, we see that the incorrect call to `outputJSON` does result in an error `TypeError: Cannot read property 'replace' of undefined at stringify...` (corresponds to [this call](https://github.com/jprichardson/node-fs-extra/blob/a149f822c387958caba7333ad2830615ab8b078f/lib/json/output-json.js#L7) being made with `undefined`). And so, the promise resulting from the call to `outputJSON` is rejected. As such, the reject branch of the `.then` is triggered, which results in our callback being executed. 

We see that in our callback, the `TypeError` is printed as it is the argument to the callback.

#### `pathExists`: another function with this behaviour
The [`pathExists` function is also universalified with `fromPromise`](https://github.com/jprichardson/node-fs-extra/blob/master/lib/path-exists/index.js#L10). As such, calling `pathExists` with only a callback has the same behaviour as `outputJSON` (i.e. the callback is executed).
```
fse.pathExists("newfile.json", callback); // newfile.json exists
> Callback executing!
> [Arguments] { '0': null, '1': true}

// incorrect call still executes the callback
fse.pathExists( callback); 
> Callback executing!
>  [Arguments] { '0': null, '1': false}
```

### Comparison to behaviour of corresponding 'fs' functions
`outputJSON` is a special case of `writeFile`, so I assume it should have similar error checking to `fs.writeFile`.
`pathExists` is checking for the existence of a file and is just a wrapper for `fs.access`, so it should have the same error checking as `fs.access`.
Both of these functions have a type error if you try and call them with just a callback.
```
let fs = require('fs');
fs.writeFile(callback); // this function expects 3 arguments, and the last must be a callback
> Uncaught:                                                                                                                                                                            
TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received undefined                                                                                                    
    at maybeCallback (fs.js:160:9)
    at Object.writeFile (fs.js:1431:14)

fs.access(callback);
Uncaught:                                                                                                                                                                            
TypeError [ERR_INVALID_CALLBACK]: Callback must be a function. Received undefined                                                                                                    
    at maybeCallback (fs.js:160:9)
    at Object.writeFile (fs.js:1431:14)
```

### Fix 
It is not enough to just add argument type checking code to the implementations of `outputJSON` and `pathExists`: because of the design of `fromPromise`, the callback will be executed regardless of the contents of these functions.
I have a proposed fix that simply involves replicating the implementation of `fromPromise` in the body of these functions, after some type checking. 
I'll make a PR soon and link it here.

emarteca:
Fixes #865 

And, adds tests for this new error checking.

### DIFF
--- a/lib/json/__tests__/output-json.test.js
+++ b/lib/json/__tests__/output-json.test.js
@@ -77,5 +77,28 @@ describe('json', () => {
         })
       })
     })
+
+    describe('> exit with error when missing or invalid arguments', () => {
+      it('passing non-string for file should result in error', done => {
+        fse.outputJson(3, {}).catch(err => {
+          assert.strictEqual(err.toString(), 'TypeError: [ERR_INVALID_ARG_TYPE] the "file" argument must be of type string')
+          done()
+        })
+      })
+
+      it('passing non-object for data should result in error', done => {
+        fse.outputJson('file.json', 'bad arg').catch(err => {
+          assert.strictEqual(err.toString(), 'TypeError: [ERR_INVALID_ARG_TYPE] the "data" argument must be of type object')
+          done()
+        })
+      })
+
+      it('passing only a callback argument should result in error', done => {
+        fse.outputJson(() => {}).catch(err => {
+          assert.strictEqual(err.toString(), 'TypeError: [ERR_INVALID_ARG_TYPE] the "file" argument must be of type string')
+          done()
+        })
+      })
+    })
   })
 })
--- a/lib/json/index.js
+++ b/lib/json/index.js
@@ -1,9 +1,8 @@
 'use strict'
 
-const u = require('universalify').fromPromise
 const jsonFile = require('./jsonfile')
 
-jsonFile.outputJson = u(require('./output-json'))
+jsonFile.outputJson = require('./output-json')
 jsonFile.outputJsonSync = require('./output-json-sync')
 // aliases
 jsonFile.outputJSON = jsonFile.outputJson
--- a/lib/json/output-json.js
+++ b/lib/json/output-json.js
@@ -8,7 +8,7 @@ async function outputJson (file, data, options = {}, callback = undefined) {
     throw TypeError('[ERR_INVALID_ARG_TYPE] the "file" argument must be of type string')
   }
   if (typeof data !== 'object') {
-    throw TypeError('[ERR_INVALID_ARG_TYPE] the "data" argument must be a JSON object')
+    throw TypeError('[ERR_INVALID_ARG_TYPE] the "data" argument must be of type object')
   }
 
   // if the options argument is a function and there is no callback argument
--- a/lib/json/output-json.js
+++ b/lib/json/output-json.js
@@ -3,10 +3,28 @@
 const { stringify } = require('jsonfile/utils')
 const { outputFile } = require('../output')
 
-async function outputJson (file, data, options = {}) {
+async function outputJson (file, data, options = {}, callback = undefined) {
+  if (typeof file !== 'string') {
+    throw TypeError('[ERR_INVALID_ARG_TYPE] the "file" argument must be of type string')
+  }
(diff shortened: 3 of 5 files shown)


RyanZim:
I'm not keen to duplicate business logic from universalify into every function body; the fact that `fs` does type checking of arguments is unusual for Node.js, but is necessary since Node core is backed by native code. I don't see the behavior of `outputJSON` as problematic personally; async functions generally transmit their errors by calling callbacks anyhow.

The `pathExists` behavior does seem to be a bug, since no error is thrown if only a callback is passed. `pathExists` should error out if no path is passed to it. I don't particularly care whether the error is thrown or given to the callback, but there should be an error. This also affects using `pathExists` as a promise; `pathExists()` should not resolve to `false`, as it does currently.

PR #869, opened 2021-02-22
Clarify docs for ensureSymlink() when destPath already exists

mwest1066:
The existing docs for `ensureSymlink()` don't specify what happens in the case where `destPath` already exists. Looking at the code it turns out that when `destFile` already exists the function exits without making any modifications, whether or not `destPath` is a symlink to `srcPath`: https://github.com/jprichardson/node-fs-extra/blob/1625838cdfc65a1bbf28ab5fa962a75805629b9c/lib/ensure/symlink.js#L26

I found this behavior surprising, because the function name `ensureSymlink()` sounds like it will ensure that the symlink `destPath -> srcPath` exists. This PR clarifies the existing behavior in the docs.

### DIFF
--- a/docs/ensureSymlink.md
+++ b/docs/ensureSymlink.md
@@ -1,6 +1,6 @@
 # ensureSymlink(srcPath, destPath[, type][, callback])
 
-Ensures that the symlink exists. If the directory structure does not exist, it is created.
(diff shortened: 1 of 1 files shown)


mwest1066:
Sorry! I just realized that #786 and #826 already address this issue.

I would love to see #826 merged, but until it is, it would be great to clarify the current behavior in the docs.

RyanZim:
Gonna try to land https://github.com/jprichardson/node-fs-extra/pull/826 soon, so closing this.

PR #882, opened 2021-04-02
BREAKING: Use fs.rm/rmSync where supported

### LINKED ISSUE #806: Recursive fs.rmdir / fs.rm
Node core has added an experimental `recursive` option to `fs.rmdir` in https://github.com/nodejs/node/pull/29168; first released in v12.10.0.

Docs: https://nodejs.org/api/fs.html#fs_fs_rmdir_path_options_callback

I'm hesitant to use this as long as it's experimental in Node core, but just opening an issue to track this.

RyanZim:
Fixes #806

Technically a breaking change since this removes the undocumented ability to pass options to remove*()

### DIFF
--- a/lib/remove/index.js
+++ b/lib/remove/index.js
@@ -1,9 +1,22 @@
 'use strict'
 
+const fs = require('graceful-fs')
 const u = require('universalify').fromCallback
 const rimraf = require('./rimraf')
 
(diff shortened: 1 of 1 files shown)


manidlou (approved):
LGTM

PR #883, opened 2021-04-02
Ensure copy filter fn is not called more than needed

### LINKED ISSUE #809: copy filter called twice
- **Operating System:** MacOS 10.14.6
- **Node.js version:** 14.1.0
- **`fs-extra` version:** 9.0.1

Hi,

Copy filter function called twice for same source and destination. I found a similar report for `emptyDir`, but couldn't see a case for `copy`.

Is this expected for some reason?

Example:

```js
// Below code outputs: 1 2, whereas I expected only 1.
const { copy } = require("fs-extra");

let i = 1;

copy("README.md", "README2.md", {
  filter: () => {
    console.log(i++);
    return true;
  },
});
```

RyanZim:
Fixes #809

We need to check filter before creating a parent directory for `dest`, in case the filter cancels the entire operation. After creating the parent (if needed), we were calling a function that rechecked the filter, which wasn't needed, and resulted in the filter fn being called twice at the root level.

With these changes, we should possibly consider renaming some functions, as it doesn't seem to make much sense anymore.

### DIFF
--- a/lib/copy-sync/__tests__/copy-sync-file.test.js
+++ b/lib/copy-sync/__tests__/copy-sync-file.test.js
@@ -104,6 +104,25 @@ describe('+ copySync() / file', () => {
       assert(fs.existsSync(destFile3))
     })
 
+    it('should not call filter fn more than needed', () => {
+      const src = path.join(TEST_DIR, 'foo')
+
+      fs.writeFileSync(src, '')
+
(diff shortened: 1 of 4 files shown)


RyanZim:
@manidlou ping!

manidlou (approved):
LGTM

PR #884, opened 2021-04-02
Clarify docs for copy*() filter

### LINKED ISSUE #843: [copy] filter is applied to directories, even though docs say it's just applied to files
Hence I can't use a filter like `file => file.endsWith('.js')` because that excludes directories that don't end with `.js`.

The docs only mention files:
> filter <Function>: Function to filter copied files. Return true to include, false to exclude. Can also return a Promise that resolves to true or false (or pass in an async function).

However when I log what the filter gets called with, it gets called with directories.

RyanZim:
Fixes #843

### DIFF
--- a/docs/copy-sync.md
+++ b/docs/copy-sync.md
@@ -9,7 +9,7 @@ Copy a file or directory. The directory can have contents.
   - `errorOnExist` `<boolean>`: when `overwrite` is `false` and the destination exists, throw an error. Default is `false`.
   - `dereference` `<boolean>`: dereference symlinks, default is `false`.
   - `preserveTimestamps` `<boolean>`: When true, will set last modification and access times to the ones of the original source files. When false, timestamp behavior is OS-dependent. Default is `false`.
-  - `filter` `<Function>`: Function to filter copied files. Return `true` to include, `false` to exclude.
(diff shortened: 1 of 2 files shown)


docs/copy-sync.md
```diff
@@ -9,7 +9,7 @@ Copy a file or directory. The directory can have contents.
   - `errorOnExist` `<boolean>`: when `overwrite` is `false` and the destination exists, throw an error. Default is `false`.
   - `dereference` `<boolean>`: dereference symlinks, default is `false`.
   - `preserveTimestamps` `<boolean>`: When true, will set last modification and access times to the ones of the original source files. When false, timestamp behavior is OS-dependent. Default is `false`.
-  - `filter` `<Function>`: Function to filter copied files. Return `true` to include, `false` to exclude.
+  - `filter` `<Function>`: Function to filter copied files/directories. Return `true` to include, `false` to exclude.
```
manidlou, line 12: I believe it would be clearer if we say `Return true to copy the item, false to ignore it.`

JPeer264: I like that proposal.

RyanZim: Updated

manidlou (approved):
LGTM

PR #888, opened 2021-04-15
Clarify move*() docs

### LINKED ISSUE #830: Add docs to explain whether `move` supports directories as destination
Given that `mv somefile.txt textfiles` will happily move a file `somefile.txt` into a dir `textfiles`, one might expect `move('somefile.txt', 'textfiles')` to work, but the docs currently do not explain what a valid `dest` is. Could the docs be amended to make it explicit whether `dest` must be a file path, or whether any path is valid?

RyanZim:
Fixes #830

### DIFF
--- a/docs/move-sync.md
+++ b/docs/move-sync.md
@@ -3,7 +3,7 @@
 Moves a file or directory, even across devices.
 
 - `src` `<String>`
-- `dest` `<String>`
+- `dest` `<String>` Path to move `src` to
 - `options` `<Object>`
   - `overwrite` `<boolean>`: overwrite existing file or directory, default is `false`.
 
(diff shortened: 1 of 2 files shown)


manidlou (commented):
@RyanZim this is ok but I think it's still not clear that we are not supporting moving a file to a directory. That's something I've seen a lot from users' comments. People think whatever they can do with `mv` command, they also can do with `fs-extra`, but we know that's not true! We need to make that clear for our users.

RyanZim:
Yeah, I couldn't figure out a way to state this more clearly; I'm not good with words that way. Any suggestions?

manidlou:
To be honest, I am not sure neither but I was thinking something like

`dest <String> when src is a file, dest must be a file and when src is a directory, dest must be a directory`

manidlou (approved):
LGTM! Thanks @RyanZim!

Issue #886, opened 2021-04-03
Drop Node v10 support


RyanZim:
Node v10 EOLs on 2021-04-30; here's a list of things we can do when we drop support for Node v10:

- `fs.promises` special handling not needed: https://github.com/jprichardson/node-fs-extra/blob/2fca5b213c86a659d0c86967cf5130c40145d2d5/lib/index.js#L20-L27 & https://github.com/jprichardson/node-fs-extra/blob/2fca5b213c86a659d0c86967cf5130c40145d2d5/lib/fs/index.js#L52-L56
- `mkdirs` can use native `fs.mkdir` with `recursive` option everywhere https://github.com/jprichardson/node-fs-extra/blob/2fca5b213c86a659d0c86967cf5130c40145d2d5/lib/mkdirs/make-dir.js#L11
- No special handling to avoid `bigint` usage https://github.com/jprichardson/node-fs-extra/blob/2fca5b213c86a659d0c86967cf5130c40145d2d5/lib/util/stat.js#L8 & https://github.com/jprichardson/node-fs-extra/blob/2fca5b213c86a659d0c86967cf5130c40145d2d5/lib/util/__tests__/stat.test.js#L10

JPeer264:
Also do not forget to update the `engines` `package.json`

https://github.com/jprichardson/node-fs-extra/blob/c8815e3ccf130422b427484007f73029075b56b6/package.json#L5-L7



If #881 is not yet implemented by this time Node10 got dropped and `.travis,yml` is still in, it should be updated as well (same with appveyor)


https://github.com/jprichardson/node-fs-extra/blob/c8815e3ccf130422b427484007f73029075b56b6/.travis.yml#L6-L7

RyanZim:
Done in https://github.com/jprichardson/node-fs-extra/pull/893, https://github.com/jprichardson/node-fs-extra/pull/890,  https://github.com/jprichardson/node-fs-extra/pull/894, & https://github.com/jprichardson/node-fs-extra/pull/895.

PR #890, opened 2021-04-22
BREAKING: Remove special handling of fs.promises

RyanZim:
Refs #886

Only for merge after Node v10 is dropped.


manidlou (approved):
LGTM

PR #891, opened 2021-04-25
Create GitHub Pages website for docs 

MuhammadWasif:
Using `/docs` folder from `master` branch to create website for docs. Hope you'll like it.
Here's how it will look: https://muhammadwasif.github.io/node-fs-extra/

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,0 +1,1 @@
+theme: jekyll-theme-minimal
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,0 +1,264 @@
+Node.js: fs-extra
+=================
+
+`fs-extra` adds file system methods that aren't included in the native `fs` module and adds promise support to the `fs` methods. It also uses [`graceful-fs`](https://github.com/isaacs/node-graceful-fs) to prevent `EMFILE` errors. It should be a drop in replacement for `fs`.
+
+[![npm Package](https://img.shields.io/npm/v/fs-extra.svg)](https://www.npmjs.org/package/fs-extra)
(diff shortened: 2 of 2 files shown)


RyanZim:
I'm personally not sure if a website improves on the existing GitHub docs much, but I'm not a hard no on this. @jprichardson @manidlou @JPeer264 Thoughts?

One thing that definitely needs to change: we can't duplicate the entire README here; we need to come up with something else.

JPeer264:
I'm usually also pretty minimalistic and think the GitHub docs do just fine, but maybe some people like it better than the GitHub docs. Furthermore it doesn't seem that it produces much more overhead, except with this README duplication where I'm also up for a change.

I never used Jekyll as I like the GitHub docs as is, but after some research I found this which could help removing this duplicated content: https://github.com/benbalter/jekyll-readme-index

MuhammadWasif:
I am using GitHub Pages to deploy `/docs` folder. It contains `_config.yml` that deploys the folder. Folder must contain an `index.md` to use as a home page. So, we need to have the README as `index.md` in `/docs` folder.
 `jekyll-readme-index` can't be used as their is no gem file and docs are deployed by GitHub Pages only.

JPeer264:
Sorry for the delayed response. Ok clearly I know too little about gem's.

I don't think it's a good idea to have that readme duplication in here. Furthermore, if this one is only available in ghpages (which makes sense to have it there), we need to update ghpages everytime manually or per automatic deploy (asking myself when we should do these updates, every master build?). I think it's way too much overhead for very little profit here.

I am closing this one here, as I also noticed from the community that people find their way to the right place in the docs. If there are other opinions on this please feel free to reopen this PR. But nonetheless, thanks @MuhammadWasif for the willingness to contribute :)

PR #881, opened 2021-04-01
Implement GH actions

RyanZim:

PR #893, opened 2021-04-29
BREAKING: Drop Node 10 support

RyanZim:
~Not needed if we merge https://github.com/jprichardson/node-fs-extra/pull/881 and kill the old CI providers.~ Will need edits; still need `engines` bump.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ os:
   - linux
   - osx
 node_js:
-  - 10
-  - 11
   - 12
   - 13
   - 14
+  - 15
+  - 16
 env: TEST_SUITE=unit
 matrix:
   exclude:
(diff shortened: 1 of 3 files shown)


RyanZim:
As usual, AppVeyor is a sloth for supporting new Node.js versions; I'll have to drop Node 16 tests on Windows for the time being. Yet another reason to switch to GH Actions.

JPeer264 (approved):
Just for my understanding, is there actually a reason why we don't test node v15 on AppVeyor?

RyanZim:
@JPeer264 AppVeyor's performance generally isn't great, so we've traditionally only tested LTS releases there, to make the build run faster. GH Actions fixes this.

PR #896, opened 2021-04-29
Move at-least-node to devDependencies

RyanZim:
After #894 & #895

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,12 +37,12 @@
   "author": "JP Richardson <jprichardson@gmail.com>",
   "license": "MIT",
   "dependencies": {
-    "at-least-node": "^1.0.0",
     "graceful-fs": "^4.2.0",
     "jsonfile": "^6.0.1",
     "universalify": "^2.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -37,12 +37,12 @@
   "author": "JP Richardson <jprichardson@gmail.com>",
   "license": "MIT",
   "dependencies": {
-    "at-least-node": "^1.0.0",
     "graceful-fs": "^4.2.0",
```
JPeer264, line 40: Not target of this PR, but should we update `graceful-fs` (and maybe all the dev dependencies too?). Just saw that there is already v`4.2.6`. Could be a nice move during our major version bump

JPeer264 (approved):

package.json
RyanZim: The `^` will already pull in the latest version; I don't see the point of bumping it explicitly.

I'm fine with bumping dev dependencies, but no need to tie that to the major bump. Feel free to jump on that whenever you have time.

JPeer264: Ah ya true my bad!

manidlou (approved):
LGTM

PR #897, opened 2021-05-01
move: do not create parent directory if it is root

### LINKED ISSUE #819: Move fails with EPERM when destination is in root of drive
- **Operating System:** Windows 10 1909
- **Node.js version:** 14.2.0
- **`fs-extra` version:** 9.0.1

Moving a file or directory where the destination is in the root of a drive (e.g. `D:\filename`) causes the operation to fail, as the move function attempts to mkdir the root of the drive.

```js
const fs = require("fs-extra");
fs.moveSync("D:\\Source.txt", "D:\\Destination\\SomeFolder"); // succeeds
fs.moveSync("D:\\Source.txt", "D:\\Destination"); // fails
```

Error is as follows:

```
Uncaught Error: EPERM: operation not permitted, mkdir 'D:\'
    at Object.mkdirSync (fs.js:940:3)
    at module.exports.makeDirSync (---\node_modules\fs-extra\lib\mkdirs\make-dir.js:101:15)
    at Object.moveSync (---\node_modules\fs-extra\lib\move-sync\move-sync.js:16:3) {
  errno: -4048,
  syscall: 'mkdir',
  code: 'EPERM',
  path: 'D:\\'
}
```

manidlou:
Fixes #819.

Do not create parent directory if it is root. We run tests for this only on windows, moving to root on unix-like systems requires sudo privileges.

### DIFF
--- a/lib/move-sync/__tests__/move-sync.test.js
+++ b/lib/move-sync/__tests__/move-sync.test.js
@@ -271,19 +271,19 @@ describe('moveSync()', () => {
   })
 
   describeIfWindows('> when dest parent is root', () => {
+    let dest
+
+    afterEach(() => fse.removeSync(dest))
+
     it('should not create parent directory', () => {
       const src = path.join(TEST_DIR, 'a-file')
-      const dest = path.join(path.parse(TEST_DIR).root, 'another-file')
+      dest = path.join(path.parse(TEST_DIR).root, 'another-file')
 
       fse.moveSync(src, dest)
 
       const contents = fs.readFileSync(dest, 'utf8')
       const expected = /^sonic the hedgehog\r?\n$/
       assert(contents.match(expected))
-      // this test is a bit special since
-      // we're moving to root, so we need to
-      // manually remove dest after we're done.
-      fse.removeSync(dest)
     })
   })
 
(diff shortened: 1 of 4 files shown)


lib/move-sync/__tests__/move-sync.test.js
```diff
@@ -268,6 +270,23 @@ describe('moveSync()', () => {
...
+
+      fse.moveSync(src, dest)
+
+      const contents = fs.readFileSync(dest, 'utf8')
+      const expected = /^sonic the hedgehog\r?\n$/
+      assert(contents.match(expected))
+      // this test is a bit special since
+      // we're moving to root, so we need to
+      // manually remove dest after we're done.
+      fse.removeSync(dest)
```
RyanZim, line 286: Should be in an `afterEach` in this `describeIfWindows` block.

lib/move/__tests__/move.test.js
```diff
@@ -263,14 +265,34 @@ describe('+ move()', () => {
...
+      fse.move(src, dest, err => {
+        assert.ifError(err)
+        fs.readFile(dest, 'utf8', (err, contents) => {
+          const expected = /^sonic the hedgehog\r?\n$/
+          assert.ifError(err)
+          assert.ok(contents.match(expected))
+          // this test is a bit special since
+          // we're moving to root, so we need to
+          // manually remove dest after we're done.
+          fse.remove(dest, done)
```
RyanZim, line 290: Same here.

RyanZim (commented):
LGTM, but one flaw with the cleanup code in the tests; they will not run properly if the tests fail.

lib/move-sync/__tests__/move-sync.test.js
manidlou: Makes sense! I will update it then.

RyanZim (approved):

Issue #898, opened 2021-05-04
Test 'ncp / error / dest-permission' / 'should return an error:' fails inside a container


dcermak:
- **Operating System:** openSUSE Tumbleweed, Fedora 33, Ubuntu
- **Node.js version:** 15, 16
- **`fs-extra` version:** git head

I am currently trying to run the unit tests of node-fs-extra in a container, but sadly the `ncp / error / dest-permission` / `should return an error:` test fails consistently with:
```
  1) ncp / error / dest-permission
       should return an error:
     Uncaught AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(err)

      at /node-fs-extra/lib/copy/__tests__/ncp/ncp-error-perm.test.js:45:7
      at /node-fs-extra/node_modules/graceful-fs/polyfills.js:247:20
      at FSReqCallback.oncomplete (node:fs:184:23)
```

How to reproduce:
```ShellSession
$ podman run --rm -it node /bin/bash
# git clone https://github.com/jprichardson/node-fs-extra.git
Cloning into 'node-fs-extra'...
remote: Enumerating objects: 5760, done.
remote: Counting objects: 100% (135/135), done.
remote: Compressing objects: 100% (102/102), done.
remote: Total 5760 (delta 56), reused 83 (delta 31), pack-reused 5625
Receiving objects: 100% (5760/5760), 1.17 MiB | 10.03 MiB/s, done.
Resolving deltas: 100% (3511/3511), done.
root@950ba24c05d3:/# cd node-fs-extra/
root@950ba24c05d3:/node-fs-extra# npm install
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated mkdirp@0.5.1: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

added 426 packages, and audited 427 packages in 15s

51 packages are looking for funding
  run `npm fund` for details

3 low severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
npm notice
npm notice New minor version of npm available! 7.10.0 -> 7.11.2
npm notice Changelog: https://github.com/npm/cli/releases/tag/v7.11.2
npm notice Run npm install -g npm@7.11.2 to update!
npm notice
[...]

PR #899, opened 2021-05-05
Skip the ncp / error / dest-permission test if running as root

dcermak:
This test tries to create a directory where the current user does not have the
permissions to do so and expects a failure. However, the root user is by design
always allowed to do that (ignoring file permissions). So we must skip the test
in these cases then.

This fixes https://github.com/jprichardson/node-fs-extra/issues/898

### DIFF
--- a/lib/copy/__tests__/ncp/ncp-error-perm.test.js
+++ b/lib/copy/__tests__/ncp/ncp-error-perm.test.js
@@ -21,7 +21,9 @@ describe('ncp / error / dest-permission', () => {
   const src = path.join(TEST_DIR, 'src')
   const dest = path.join(TEST_DIR, 'dest')
 
-  if (os.platform().indexOf('win') === 0) return
+  // when we are root, then we will be able to create the subdirectory even if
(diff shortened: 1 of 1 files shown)


lib/copy/__tests__/ncp/ncp-error-perm.test.js
```diff
@@ -21,7 +21,9 @@ describe('ncp / error / dest-permission', () => {
   const src = path.join(TEST_DIR, 'src')
   const dest = path.join(TEST_DIR, 'dest')
 
-  if (os.platform().indexOf('win') === 0) return
+  // when we are root, then we will be able to create the subdirectory even if
+  // we don't have the permissions to do so, so no point in running this test
+  if ((os.platform().indexOf('win') === 0) || (os.userInfo().uid === 0)) return
```
RyanZim, line 26: Unneeded parenthesis
```suggestion
  if (os.platform().indexOf('win') === 0 || os.userInfo().uid === 0) return
```

RyanZim (approved):

manidlou (approved):

PR #906, opened 2021-06-03
Add copy on write support

In-line:


RyanZim:
I'm not opposed to the idea of adding this, but if we're going to do this, it needs to be mirrored for `copySync`. Also, we should have a test that uses this flag; not sure that we can actually test that it's working correctly, but we should at least exercise the happy path here.

lib/copy/copy.js
```diff
@@ -95,7 +95,9 @@ function mayCopyFile (srcStat, src, dest, opts, cb) {
 }
 
 function copyFile (srcStat, src, dest, opts, cb) {
-  fs.copyFile(src, dest, err => {
+  // Enable copy on write on supported systems
+  const mode = opts.copyOnWrite ? fs.constants.COPYFILE_FICLONE : 0;
```
RyanZim, line 99: linter
```suggestion
  const mode = opts.copyOnWrite ? fs.constants.COPYFILE_FICLONE : 0
```

RyanZim:
Closing as abandoned; if someone wants to pick this up, please open a new PR.

PR #922, opened 2021-09-21
Refactor folder structure

RyanZim:
Just making `lib/` a bit more organized; legacy decisions have influenced the current structure, no reason for it to be that way anymore.


manidlou (approved):
LGTM

PR #928, opened 2021-11-12
Return promise in outputFile

estherkim:
Without returning, calling `await outputFile` could continue without waiting for the file to be written.

Also updates a test.


RyanZim:
The test updates are good; however, the return statement is not needed. We're not handling promises there; everything is callback-based. universalify adds the promise support.

estherkim:
Ah missed that, thanks.

Issue #926, opened 2021-10-16
TypeError: Cannot read property 'name' of undefined


hepcat72:
- **Operating System:** Raspbian
- **Node.js version:** v14.18.1
- **`fs-extra` version:**:
```
>sudo npm ls -g fs-extra
/usr/local/lib
└─┬ homebridge@1.3.5
  └── fs-extra@10.0.0
```
I hope this is the right place to report this issue.  I see this error in my homebridge installation (which installs fs-extras as a dependency I believe) constantly in my homebridge error log:
```
TypeError: Cannot read property 'name' of undefined
    at exports.fromCallback (/usr/lib/node_modules/homebridge/node_modules/universalify/index.js:15:26)
    at Object.<anonymous> (/usr/lib/node_modules/homebridge/node_modules/fs-extra/lib/fs/index.js:57:27)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Function._load (/usr/lib/node_modules/pm2/node_modules/@pm2/io/src/metrics/httpMetrics.ts:190:35)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Object.<anonymous> (/usr/lib/node_modules/homebridge/node_modules/fs-extra/lib/index.js:5:6)
```
What can I do to address this?

RyanZim:
Are you running a stock version of Node.js. That error indicates that `require('fs').realpath.native` is undefined on your system

hepcat72:
This issue appears to be restricted to when you start homebridge using pm2.  I have worked around the issue by switching to hb-service.

roman-teamsystems:
Hi  @hepcat72.

I have the same issue. Did you solve yours somehow?

hepcat72:
No, I did not, but only because I simply haven't tried to figure it out. I just worked around the issue by switching from pm2 to hb-service to manage my homebridge process, though I am also having issues with node-red running via pm2. I can start it up, but after a reboot, it doesn't run and I have to start it manually. I don't know if it's related to fs* though - and I discovered that I had 2 competing pm2 installations (root vs pi). I uninstalled the root one and crossed my fingers, but I haven't checked with a reboot since then.

[...]

Issue #915, opened 2021-07-29
Universalify 2.0.0 breaks `fs.realpath`


behos:
- **Operating System:**
- **Node.js version:**
- 16.0.0
- **`fs-extra` version:**
- 9.1.0

Hi,

`fs-extra` uses `universalify` version `2.0.0` for creating `Promises` for callback methods including the built-in `fs.*` ones. In https://github.com/RyanZim/universalify/pull/13, the signature of the wrapper changed such that any non-null value for `err` would be considered a failure. Unfortunately, `fs.realpath` uses `false` as the error value when the call is successful, and this propagates as an exception in the promise.

I have raised this issue here: https://github.com/RyanZim/universalify/issues/16 but while it's being fixed, should the `universalify` dependency be rolled back?

RyanZim:
I am unable to reproduce `fs.realpath` passing `false` as the error value on Node.js 16.0.0; what's your OS?

behos:
You are right, I was running node through `bazel` which had its own version of node configured which really old. It returns null now so this issue does not apply :)

Issue #909, opened 2021-06-21
Incompatible with Nexe packaging  tool while using fs-extra 10.0.0


donthijackthegit:
- **Operating System:**Ubuntn 20.04.1
- **Node.js version:** v12.20.2/v14.17.1

- **`fs-extra` version:** 10.0.0

I am using Nexe to package my service into an executable file. Compilation was success, but when I run the app, I got below error message:
```
 }, 'name', { value: fn.name })
                         ^

TypeError: Cannot read property 'name' of undefined
    at exports.fromCallback (/***/dist/nexe/node_modules/universalify/index.js:15:26)
    at Object.<anonymous> (/***/dist/nexe/node_modules/fs-extra/lib/fs/index.js:57:27)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:14)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at Object.<anonymous> (/***/dist/nexe/node_modules/fs-extra/lib/index.js:5:6)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)

```
I switch back to fs-extra 9.1.0 everything works fine.

Cannot reproduce if run directly via node.

RyanZim:
This is a bug with Nexe, it looks like they don't support [`fs.realpath.native`](https://nodejs.org/api/fs.html#fs_fs_realpath_native_path_options_callback).

Issue #144, opened 2015-06-27
tests failing


futagoza:
# Implementation
Your new implementation of running the tests is okay if the module being tested isn't under some folder called `node_modules` some where in its pathname, but if it is, 0 tests are found.
It runs fine with this code: https://shrib.com/fs-extra-test.js-fix?v=pr

# Copy, move and ncp
I'm not sure these other errors are related to me running on Windows 8.1 x64, but thought I should put them up: https://shrib.com/GsntR4KJ?v=pr

jprichardson:
Your first point about the tests not running, I'm not clear about the case where they're not?

Regarding the tests failing, that's because you pulled the master branch (development). There's one issue that I believe to be a bug in io.js. More details here: https://github.com/jprichardson/node-fs-extra/pull/141.

futagoza:
Your `tests.js` file works if its located in `C:/dev/node-fs-extra`, finding 60+ tests, but fails with your code if the file is located in `C:/dev/efe/node_modules/fs-extra`, finding 0 tests.

In [my edited `file.js`](https://shrib.com/fs-extra-test.js-fix?v=pr) I have removed the line that looks for node_modules in the file path, also setting the walk method to look in `./lib` only. This works fine and ensures that it wont look inside the modules own `node_modules` directory.

If its really required to search multiple folders to add tests to mocha, then you should wrap the walker ([see example: https://shrib.com/fs-extra-test.js-modified?v=pr](https://shrib.com/fs-extra-test.js-modified?v=pr)). This should avoid looking in folders that are in `node_modules`, unless of course there's a reason to scan 5000+ file names ;)

futagoza:
@jprichardson any update on this? Right now I've got a edited cloned repo with the changes I've mentioned above so that I can optionally run the fs-extra tests (correctly, see above note about 'node_modules') along with mine

jprichardson:
Sure, please submit a PR and I'll accept it. But note, it should find > 100 tests in either scenario.

futagoza:
Ok, I'll do that later on today when I get home

futagoza:
Sorry for the delay, I got wrapped up in a problem.

I cloned from [#ff2d36](https://github.com/jprichardson/node-fs-extra/commit/ff2d362f059d4f28c0b866cfaef675816ad4b9c2), which is 5 commits ago as of now.

The modified `test.js` is only finding 43 tests, I can't find the other 50+
Are you sure there are 100+ tests?

[...]

PR #936, opened 2022-01-05
Revert "Remove check for fs.realpath.native support, since it's every where (#887)"

geralfonso:
This reverts commit f4a880d29c3a699c0695755fe8533d786c55f65f.

Operating System: Linux/mac

Node.js version: v12.22.6/v16.13.1

fs-extra version: 10.0.0

The simple check can avoid countless headaches especially for large codebases that need fs-extra version 10. It would be awesome if we could get it back because the issue is happening just with the import statement.

Fixes #

- [TypeError: Cannot read property 'name' of undefined](https://github.com/jprichardson/node-fs-extra/issues/926)
- [Universalify 2.0.0 breaks fs.realpath ](https://github.com/jprichardson/node-fs-extra/issues/915)
- [Incompatible with Nexe packaging tool while using fs-extra 10.0.0](https://github.com/jprichardson/node-fs-extra/issues/909)

### DIFF
--- a/lib/fs/index.js
+++ b/lib/fs/index.js
@@ -54,7 +54,6 @@ Object.assign(exports, fs)
 api.forEach(method => {
   exports[method] = u(fs[method])
 })
-exports.realpath.native = u(fs.realpath.native)
 
 // We differ from mz/fs in that we still ship the old, broken, fs.exists()
 // since we are a drop-in replacement for the native module
@@ -116,4 +115,9 @@ if (typeof fs.writev === 'function') {
       })
     })
   }
+}
+
+// fs.realpath.native only available in Node v9.2+
+if (typeof fs.realpath.native === 'function') {
+  exports.realpath.native = u(fs.realpath.native)
 }


RyanZim:
The question is why you don't have `fs.realpath.native`?

geralfonso:
@RyanZim Definitely you are right!

Look I was able to replicate it: [sandbox](https://codesandbox.io/s/priceless-galois-n4oud?file=/src/index.js)

Interestingly, when we require `fs-extra` before, in this case `aws-xray-sdk-core`, it works perfectly. But when requiring it after, `fs.realpath.native` no longer exists and throws the error. With native `fs` it always works.

It seems that `fs.realpath.native` is undefined from `graceful-fs`

RyanZim:
AFAIK, `graceful-fs` has always supported it; so that doesn't make much sense.

geralfonso:
I mean when the error happens It seems that ` fs.realpath.native` already comes as `undefined` from `graceful-fs`

RyanZim:
Yes, but why, when `graceful-fs` doesn't normally do that?

geralfonso:
I traced the issue to `async-listener` package. I have pushed a fix to that repo: [#144](https://github.com/othiym23/async-listener/pull/144).

Essentially they are wrapping `fs.realpath` but they are not copying the original function object properties to the wrapper. So we're getting the wrapper with missing properties. 

It would be great if we could keep the check or a fallback (i.e. realpathSync.native) to prevent collision with other libraries that are using `async-listener` as dependency.

matanarbel:
I have the same issue, and I also traced it back to `async-listener`.
@geralfonso - you're right that it's not an `fs-extra` issue, but why not simply add the check back?
It doesn't harm any functionality of fs-extra, and it'll make the usage more stable when using other old packages (in my case, `async-listener` is a dependency of a dependency of a dependency... about 5 levels of dependencies, so I can't really change it).

It'd be really helpful if you'd bring back the check!

PR #938, opened 2022-01-07
chore: update graceful-fs to 4.2.9

bluelovers:
update graceful-fs to 4.2.9 for support throwIfNoEntry for statSync


RyanZim:
The version is already defined with a `^`, so it should pull the latest version automatically.

PR #941, opened 2022-02-18
chore: "sideEffects": false

jacob-ebey:
Add `"sideEffects": false` to package.json to allow frameworks that mix client and server code in the same files to tree-shake out fs-extra from client bundles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,5 +62,6 @@
     "test-find": "find ./lib/**/__tests__ -name *.test.js | xargs mocha",
     "test": "npm run lint && npm run unit",
     "unit": "nyc node test.js"
-  }
+  },
+  "sideEffects": false
 }


RyanZim:
I've never heard of this key; what bundlers use it? Where's it documented?

jacob-ebey:
All of them:
https://esbuild.github.io/api/#conditionally-injecting-a-file
https://webpack.js.org/guides/tree-shaking/ & https://webpack.js.org/guides/tree-shaking/
https://www.npmjs.com/package/@rollup/plugin-node-resolve/v/13.0.6#user-content-ignoresideeffectsforroot

jacob-ebey:
🎉🥳 Thank you! When can we look forward to a release?

RyanZim:
Published v10.0.1

PR #948, opened 2022-03-16
Allow passing undefined options to move()

### LINKED ISSUE #947: Allow passing undefined options to move()
- **Operating System:**macos
- **Node.js version:**14.18.3
- **`fs-extra` version:**10.0.1

The follow code:

```
const { move } = require("fs-extra");

move("/some/path/that/exists", "some/path/that/doesnt/exist", undefined)
  .then(() => console.log("success")
  .catch(console.error);
```

produces the following type error: `TypeError: Cannot read property 'overwrite' of undefined`

The error seems to be from the handling of `optional callback` and `optional options` on line 12-15 of `lib/move/move.js`

Nokel81:
fixes #947 

I thought about adding a test but then couldn't find any tests of a similar nature within the repo so I didn't. I can if it is thought to be useful.

### DIFF
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -340,13 +340,7 @@ describe('+ move()', () => {
 
       fse.move(SRC_DIR, DEST_DIR, undefined)
         .then(() => done())
-        .catch(error => {
-          if (error instanceof TypeError) {
-            done()
-          } else {
-            done(error)
-          }
-        })
+        .catch(done)
     })
   })
 
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -52,6 +52,65 @@ describe('+ move()', () => {
 
   afterEach(done => fse.remove(TEST_DIR, done))
 
+  describe('> invalid arguments', () => {
+    it('should throw a type error if src is not a string', (done) => {
+      try {
+        fse.move(15, '', {}, () => ({}))
+        done('Failed to throw a type error')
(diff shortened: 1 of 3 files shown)


lib/move/move.js
```diff
@@ -14,6 +14,8 @@ function move (src, dest, opts, cb) {
     opts = {}
   }
 
+  opts = opts || {};
```
RyanZim, line 17:
```suggestion
  opts = opts || {}
```

RyanZim, line 17:
```suggestion
  opts = opts || {}
```

Nokel81:
I have fixed the implementation to work with the existing tests and decided to add a unit test to cover this use case as well.

lib/move/__tests__/move.test.js
```diff
@@ -331,6 +331,25 @@ describe('+ move()', () => {
...
+
+      fse.move(SRC_DIR, DEST_DIR, undefined)
+        .then(() => done())
+        .catch(error => {
+          if (error instanceof TypeError) {
+            done()
+          } else {
+            done(error)
+          }
+        })
```
RyanZim, line 349: Currently, the test would pass if a `TypeError` was thrown; fix:
```suggestion
        .catch(done)
```

RyanZim:
Tests are failing; haven't looked into it in detail.

RyanZim:
Created https://github.com/jprichardson/node-fs-extra/pull/955; which doesn't have the extra parameter validation, and has simpler tests.

PR #954, opened 2022-04-08
refactor: replace console.warn with process.emitWarning

peteriman:
To have consistent warnings from https://github.com/jprichardson/node-fs-extra/pull/953#issuecomment-1092663914 by @JPeer264 
> Yes `process.emitWarning` sounds also like a great option to add (just my opinion on it). However, a separate PR would be nice if you want to go the extra mile as we squash our PRs. @manidlou @RyanZim @jprichardson any objections?

Prints out via `process.emitWarning` in the following format:

image

### DIFF
--- a/lib/copy/copy-sync.js
+++ b/lib/copy/copy-sync.js
@@ -17,8 +17,11 @@ function copySync (src, dest, opts) {
 
   // Warn about using preserveTimestamps on 32-bit node
   if (opts.preserveTimestamps && process.arch === 'ia32') {
-    console.warn(`fs-extra: Using the preserveTimestamps option in 32-bit node is not recommended;\n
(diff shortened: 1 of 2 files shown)


RyanZim:
Should we consider this: https://nodejs.org/api/process.html#avoiding-duplicate-warnings?

peteriman:
I'm not exactly sure, as my motive was to remain the behavior of the `console.warn` which warns duplicatedly if called multiple times.

Should we change it to use  https://nodejs.org/api/process.html#avoiding-duplicate-warnings instead, so it only warns on the first time?

What do you guy think?
@RyanZim @JPeer264 @manidlou @jprichardson

JPeer264 (approved):
I am fine with either. Thanks for the changes and your efforts :)

PR #955, opened 2022-04-09
Allow passing undefined opts to move()

### LINKED ISSUE #947: Allow passing undefined options to move()

RyanZim:
Fixes #947

Closes #948

Will merge in a few days if no objections.

### DIFF
--- a/lib/__tests__/promise.test.js
+++ b/lib/__tests__/promise.test.js
@@ -10,7 +10,6 @@ const methods = [
   'ensureFile',
   'ensureDir',
   'mkdirs',
-  'move',
   'readJson',
   'readJSON',
   'remove'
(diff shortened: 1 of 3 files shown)


Nokel81:
Sounds good.

PR #971, opened 2022-10-21
Run copy*() filter before running fs.stat() on items

### LINKED ISSUE #965: [Bug] [copy] filtered-out items get statted, leading to unrecoverable exception
- **Operating System:** Windows 10 Pro
- **Node.js version:** 16.15.0
- **`fs-extra` version:** 10.1.0

### Reproduction
Try to `copy(src, dest, { filter }`, with:
+ `src` being a valid and accessible source directory, in my case a secondary drive's root
+ `src` having a child directory that is not accessible, in my case the `System Volume Information` directory that is owned by the OS and inaccessible to users
+ `dest` being a valid and accessible destination directory, in my case an empty directory on another drive
+ `filter` being a valid filter function that excludes the inaccessible directory, e.g. `const filter = (fp)=>!fp.includes("System Volume Information")`

### Observed behavior
```sh
Error: EPERM: operation not permitted, lstat 'B:\System Volume Information'
```

### Expected behavior
The copy operation proceeds without trying to access the filtered-out directory.

### Additional information
During debugging, I tracked the issue to this line:
https://github.com/jprichardson/node-fs-extra/blob/0220eac966d7d6b9a595d69b1242ab8a397fba7f/lib/copy/copy.js#L181

The way I understand the code, `src` and `dest` are `stat`ed before running the filter. I think that the filter should run before any attempt is made to stat any files.

If there are more important reasons why that should not happen, e.g. to enable #844, there should at least be a mechanism to recover from errors on a per-file basis in order to allow the copy operation to continue.

As it stands, I will have to work around this by giving a filtered set of files to `copy` 😢

RyanZim:
Fixes #965 cc @c-vetter

Exact approach in https://github.com/nodejs/node/pull/44922 didn't seem appropriate.

### DIFF
--- a/lib/copy/copy-sync.js
+++ b/lib/copy/copy-sync.js
@@ -26,18 +26,9 @@ function copySync (src, dest, opts) {
 
   const { srcStat, destStat } = stat.checkPathsSync(src, dest, 'copy', opts)
   stat.checkParentPathsSync(src, srcStat, dest, 'copy')
-  return handleFilterAndCopy(destStat, src, dest, opts)
-}
-
-function handleFilterAndCopy (destStat, src, dest, opts) {
   if (opts.filter && !opts.filter(src, dest)) return
   const destParent = path.dirname(dest)
   if (!fs.existsSync(destParent)) mkdirsSync(destParent)
-  return getStats(destStat, src, dest, opts)
-}
-
-function startCopy (destStat, src, dest, opts) {
-  if (opts.filter && !opts.filter(src, dest)) return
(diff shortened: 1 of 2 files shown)


thoqbk:
> Exact approach in https://github.com/nodejs/node/pull/44922 didn't seem appropriate.
@RyanZim do you mean this fix [44922](https://github.com/nodejs/node/pull/44922) has issue?

I have another one for cpSync [45143](https://github.com/nodejs/node/pull/45143)

PR #973, opened 2022-10-21
Remove sketchy citebite links in documentation

leafac:
I tried to follow these links, and they led me to really sketchy websites that have nothing to do with Node.js at all.


RyanZim (approved):
Yep, looks like the link got taken over. Boy, that's an old link; was added in #45.

leafac:
@RyanZim Thank you very much 😁

PR #974, opened 2022-10-25
ESM support

RyanZim:
Depends on  #970

Review commit-by-commit.

Breaking change, as it switches from `main` to `exports` in `package.json`, outlawing all subpath imports except those explicitly specified.

Fixes https://github.com/jprichardson/node-fs-extra/issues/746

### DIFF
--- a/lib/esm.mjs
+++ b/lib/esm.mjs
@@ -1,0 +1,68 @@
+import _copy from './copy/index.js'
+import _empty from './empty/index.js'
+import _ensure from './ensure/index.js'
+import _json from './json/index.js'
+import _mkdirs from './mkdirs/index.js'
+import _move from './move/index.js'
+import _outputFile from './output-file/index.js'
+import _pathExists from './path-exists/index.js'
+import _remove from './remove/index.js'
+
+// NOTE: Only exports fs-extra's functions; fs functions must be imported from "node:fs" or "node:fs/promises"
+
+export const copy = _copy.copy
+export const copySync = _copy.copySync
+export const emptyDirSync = _empty.emptyDirSync
+export const emptydirSync = _empty.emptydirSync
+export const emptyDir = _empty.emptyDir
+export const emptydir = _empty.emptydir
+export const createFile = _ensure.createFile
+export const createFileSync = _ensure.createFileSync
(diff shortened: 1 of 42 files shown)


package.json
```diff
@@ -51,16 +51,20 @@
     "read-dir-files": "^0.1.1",
     "standard": "^16.0.3"
   },
-  "main": "./lib/index.js",
+  "exports": {
+    ".": "./lib/index.js",
+    "./esm": "./lib/esm.mjs"
+  },
```
privatenumber, line 57: Have you considered using export conditions so it automatically exports the ESM version when `import` is used instead of `require`?

```
"exports": {
	".": {
		"require": "./lib/index.js",
		"import": "./lib/esm.mjs"
	},
	"./esm": "./lib/esm.mjs"
}
```

I added the `./esm` in case, but this way, you may not even need the `./esm` endpoint:

```
"exports": {
	"require": "./lib/index.js",
	"import": "./lib/esm.mjs"
}
```

lib/esm.mjs
```diff
@@ -0,0 +1,68 @@
...
+  ..._copy,
+  ..._empty,
+  ..._ensure,
+  ..._json,
+  ..._mkdirs,
+  ..._move,
+  ..._outputFile,
+  ..._pathExists,
+  ..._remove
+}
```
privatenumber, line 68: You may also be able to release this as a feature by moving this file to the root, and moving it back and adding the export map change in the next major.

package.json
RyanZim: I don't want to do that, because the ESM version doesn't export normal `fs` methods like the CJS version does. Since it's a different feature set, it's best to be explicit. Otherwise, I'd definitely do this.

lib/esm.mjs
RyanZim: I'm going to be releasing a major version for other reasons anyhow, so no reason to release as a feature.

package.json
privatenumber: Why the regression? Sounds it can be kept with:
```ts
export * from 'fs';
```

RyanZim: See https://github.com/jprichardson/node-fs-extra/issues/746#issuecomment-885076808

Issue #918, opened 2021-08-10
Port copy fix from Node.js core


RyanZim:
Just opening an issue for https://github.com/nodejs/node/pull/39372#discussion_r670744165 so this doesn't get forgotten.

RyanZim:
@manidlou FYI, will be working on pushing a new major release in the coming days with ESM support (refs https://github.com/jprichardson/node-fs-extra/issues/746#issuecomment-923250293); any chance you could tackle this soon?

rxliuli:
It seems not difficult to support esm/cjs at the same time through the build tool, why not do this?

RyanZim:
@rxliuli We are planning to support both CJS and ESM; due to some special considerations, we can't just use an out of the box solution. See the issue for more details.

RyanZim:
There's a real bug here, but I'm not sure if @bcoe fixed it correctly; see https://github.com/nodejs/node/pull/39372#discussion_r1001179295

PR #976, opened 2022-10-25
Remove unneeded and buggy stats check

RyanZim:
As per https://github.com/nodejs/node/pull/39372#discussion_r1001179295; haven't gotten a response from @bcoe there yet

Resolves https://github.com/jprichardson/node-fs-extra/issues/918

Hope this logic is right.

### DIFF
--- a/lib/copy/copy-sync.js
+++ b/lib/copy/copy-sync.js
@@ -154,7 +154,7 @@ function onLink (destStat, src, dest, opts) {
     // prevent copy if src is a subdir of dest since unlinking
     // dest in this case would result in removing src contents
     // and therefore a broken symlink would be created.
-    if (fs.statSync(dest).isDirectory() && stat.isSrcSubdir(resolvedDest, resolvedSrc)) {
+    if (stat.isSrcSubdir(resolvedDest, resolvedSrc)) {
       throw new Error(`Cannot overwrite '${resolvedDest}' with '${resolvedSrc}'.`)
     }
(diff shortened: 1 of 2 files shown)


RyanZim:
@manidlou @bcoe bump

RyanZim:
I have been trying to write a test case for this; I cannot get to a place where this condition is ever reached. Do we need this at all? Or what am I missing?

manidlou:
This is basically a wild edge case where we're trying to avoid broken copy when src is a link and its resolved path is inside the dest directory. We already have this test

https://github.com/jprichardson/node-fs-extra/blob/e6a95058c930953113177c9518f57e83cace3e79/lib/copy/__tests__/copy-prevent-copying-into-itself.test.js#L282-L296

what's wrong with this test?

RyanZim:
@manidlou The problem is that that test doesn't actually exercise this condition. We'd expect the error to be of the format:

    Cannot overwrite '${resolvedDest}' with '${resolvedSrc}'.

Instead, if you insert a log statement in that test, the actual error thrown is:

    Error: Cannot overwrite directory '/tmp/fs-extra/copy-prevent-copying-into-itself/dest' with non-directory '/tmp/fs-extra/copy-prevent-copying-into-itself/src-symlink'.

RyanZim:
I'm going to merge this as-is, without an explicit test, as it's certainly better than what we have now. However, we should still look into properly testing this later.

manidlou:
Thank you @RyanZim! That makes sense.

Issue #979, opened 2022-11-29
Cannot find module 'fs-extra/esm' or its corresponding type declarations.ts(2307)


mtdvlpr:
Both importing from fs-extra and fs-extra/esm doesn't work anymore in v11.0.0. I tried the following lines in my fs.ts file:

```
import { ensureDirSync, ensureFileSync, removeSync } from 'fs-extra'
import { ensureDirSync, ensureFileSync, removeSync } from 'fs-extra/esm'
```

The top line does work in v10.1.0. Am I doing something wrong?

- **Operating System:** Ubuntu 22.04.1 LTS
- **Node.js version:** v16.18.0
- **`fs-extra` version:** v11.0.0

RyanZim:
The fact that the first line worked in v10.1.0 indicates your typescript transpiler is doing some kind of extra compatibility layer here, which somehow was broken in v11. I'm not a typescript guy, so I can't say offhand what's gone wrong here. Are you running the latest version of TS? If all else fails, a minimal example repo will probably be useful here.

caugner:
@RyanZim I just had a quick skim through https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html and found this:

> One more wrinkle, TypeScript does not (yet) behave with exports. So you need to include the legacy `module` and `main` properties for TypeScript. The `main` property points to the CJS entry point and the module property points to the ESM entry.

Based on that, this should work:

```json
{
  "main": "./lib/index.js",
  "module": "./lib/esm.mjs",
  "exports": {
    "import": "./lib/esm.mjs",
    "require": "./lib/index.js"
  }
}
```

RyanZim:
Furthermore, your proposed change would be a breaking change; since the ESM version is not compatible with the regular version of fs-extra, we can't just ship it to all `import` users.

caugner:
> This is fixed in TypeScript v4.7

We're on TypeScript 4.9.

> Furthermore, your proposed change would be a breaking change

Not sure why you just close my **draft** PR, but okay. 🤷

RyanZim:
You may have read it already, but just to make it abundantly clear, `fs-extra/esm` is not just an ESM version of `fs-extra`: https://github.com/jprichardson/node-fs-extra#esm

I don't use TS, so perhaps I'm misunderstanding, but from the TS v4.7 release notes, it looks like they added support for `exports`. What am I missing?

> Not sure why you just close my draft PR, but okay. shrug

[...]

PR #980, opened 2022-11-29
fix(package.json): add main/module, update exports

caugner:
Fixes https://github.com/jprichardson/node-fs-extra/issues/979.


RyanZim:
Closing as per https://github.com/jprichardson/node-fs-extra/issues/979#issuecomment-1331304800.

PR #981, opened 2022-11-29
fix: add package main field for TypeScript support

### LINKED ISSUE #979: Cannot find module 'fs-extra/esm' or its corresponding type declarations.ts(2307)

mtdvlpr:
Fixes #979

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,8 @@
     "read-dir-files": "^0.1.1",
     "standard": "^17.0.0"
   },
+  "main": "./lib/index.js",
+  "module": "./lib/esm.mjs",
   "exports": {
     ".": "./lib/index.js",
     "./esm": "./lib/esm.mjs"
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "standard": "^17.0.0"
   },
   "main": "./lib/index.js",
-  "module": "./lib/esm.mjs",
   "exports": {
     ".": "./lib/index.js",
     "./esm": "./lib/esm.mjs"


caugner (approved):
See also: https://github.com/jprichardson/node-fs-extra/pull/980

Tested locally, and resolves the issue with TypeScript.

caugner:
It doesn't resolve the issue with `import from "fs-extra/esm"` though. 🤔

mtdvlpr:
It doesn't have to, right? Importing from fs-extra will get the esm version if I'm not mistaken.

caugner:
> It doesn't have to, right? Importing from fs-extra will get the esm version if I'm not mistaken.

If I understand @RyanZim's intention correctly, the separate `fs-extra/esm` export aims to avoid confusion. Adding the "module" field reintroduces that confusion, because you could then import from `fs-extra` and get the ESM module.

We could just add the `main` field for now, and that alone would resolve TypeScript issue.

package.json
```diff
@@ -51,6 +51,8 @@
     "read-dir-files": "^0.1.1",
     "standard": "^17.0.0"
   },
+  "main": "./lib/index.js",
+  "module": "./lib/esm.mjs",
```
caugner, line 55:
```suggestion
  "main": "./lib/index.js",
```

caugner (changes requested):
A more conservative solution is to omit the `module` field for now, because the author does not intend to export the ESM module at `fs-extra`.

package.json
mtdvlpr: Alright, but then it's impossible to get the esm module in TypeScript.

caugner: The solution for the ESM export (`import fse from fs-extra/esm`) might be to add a type declaration file, and replace the `./esm` entry in `exports` with an object that has two keys `import` and `types`.

Background on why the ESM export on `fs-extra` itself is undesired is here: https://github.com/jprichardson/node-fs-extra/pull/974#discussion_r1027022282

caugner: PS: I'm not actually sure what the advantage of the ESM module is (in this context). Afaik you can always import the CommonJS exports, even if you're in the context of a pure ESM package.

caugner (approved):
Re-tested and still resolves the issue when using `import fs from "fs-extra";` in TypeScript.

RyanZim:
I removed the `main` field, since it wasn't needed for backward compatibility with old Node versions, as we don't support Node versions old enough not to support `exports`. However, since it seems TS doesn't like this (potentially a bug? see https://github.com/jprichardson/node-fs-extra/issues/979#issuecomment-1331404978), it seems wise to add it back for now.

> It doesn't resolve the issue with import ... from "fs-extra/esm" though. 

I think that's fine; `main` is only needed for TS support, and if you're using TS, you already have named import support; there's no reason to use `fs-extra/esm` in TS.

@caugner is correct; we don't want to add `module` here, as that would be confusing.

RyanZim (approved):

RyanZim:
Published in v11.1.0 :tada:

PR #986, opened 2022-12-10
feat: support for esm named imports

rxliuli:
motivation

There are still problems with esm support at present, and named imports are not really available. For example, the following code will report an error when using the latest version of fs-extra

```ts
import { readdir } from 'fs-extra'
import fsExtra from 'fs-extra'
import { fileURLToPath } from 'url'
import path from 'path'

const __filename = fileURLToPath(import.meta.url)
const __dirname = path.dirname(__filename)
console.log(await readdir(__dirname))
console.log(await fsExtra.readdir(__dirname))
```

This PR addresses the following issues

- Named imports and default imports support in esm
  ```ts
  import { readdir } from 'fs-extra'
  import fsExtra from 'fs-extra'
  import { fileURLToPath } from 'url'
  import path from 'path'
  
  const __filename = fileURLToPath(import.meta.url)
  const __dirname = path.dirname(__filename)
  console.log(await readdir(__dirname))
  console.log(await fsExtra.readdir(__dirname))
  ```
- named import and default import support in cjs
  ```ts
  import { readdir } from 'fs-extra'
  import fsExtra from 'fs-extra'
  const { readdir: readdirCjs } = require('fs-extra')
  const fsExtraCjs = require('fs-extra')
  ;(async () => {
    console.log(await readdir(__dirname))
    console.log(await readdirCjs(__dirname))
    console.log(await fsExtra.readdir(__dirname))
    console.log(await fsExtraCjs.readdir(__dirname))
  })()
  ```
- Properly support the use in ts, no longer use `fs-extra/esm`

For specific test methods, see: https://github.com/rxliuli/fs-extra-demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,23 +59,23 @@ const fse = require('fs-extra')
 
 ### ESM
 
-There is also an `fs-extra/esm` import, that supports both default and named exports. However, note that `fs` methods are not included in `fs-extra/esm`; you still need to import `fs` and/or `fs/promises` seperately:
+There is also an `fs-extra` import, that supports both default and named exports. However, note that `fs` methods are not included in `fs-extra`; you still need to import `fs` and/or `fs/promises` seperately:
 
 ```js
 import { readFileSync } from 'fs'
 import { readFile } from 'fs/promises'
-import { outputFile, outputFileSync } from 'fs-extra/esm'
+import { outputFile, outputFileSync } from 'fs-extra'
 ```
 
 Default exports are supported:
 
 ```js
 import fs from 'fs'
-import fse from 'fs-extra/esm'
+import fse from 'fs-extra'
 // fse.readFileSync is not a function; must use fs.readFileSync
 ```
 
-but you probably want to just use regular `fs-extra` instead of `fs-extra/esm` for default exports:
+but you probably want to just use regular `fs-extra` instead of `fs-extra` for default exports:
 
 ```js
 import fs from 'fs-extra'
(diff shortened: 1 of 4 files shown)


RyanZim:
This approach doesn't work for a couple of reasons:

- `fs.lchown` is not available on some Linuxes
- This doesn't allow us to export functions that are only available in some Node.js versions (not the case now, but has been in the past and likely will be in the future).

For more details, see discussion on https://github.com/jprichardson/node-fs-extra/issues/746 & https://github.com/jprichardson/node-fs-extra/pull/854

rxliuli:
> This approach doesn't work for a couple of reasons:
> 
> * `fs.lchown` is not available on some Linuxes
> * This doesn't allow us to export functions that are only available in some Node.js versions (not the case now, but has been in the past and likely will be in the future).
> 
> For more details, see discussion on #746 & #854

What I don't understand is why the export interface of its cjs version is different from that of esm version. In cjs, all the methods of fs itself can be used, but in esm export, it is not included.
 I understand that fs may have different exports between nodejs versions, while cjs always forwards and exposes all, and esm must be explicitly specified, but this can be solved by version, for example, different nodejs versions are required for different major versions.
 Speaking of ts support, if you do use another export point fs-extra/esm, even if the types are fixed, this can easily be mis-imported, because you have to take care to avoid importing from fs-extra, which can be annoying

RyanZim:
> this can be solved by version, for example, different nodejs versions are required for different major versions.

It's not practical to manage multiple major version release lines for different Node versions. This also still doesn't solve the obstacle with `fs.lchown`.

lvjiaxuan:
Feel sad for this.

Issue #988, opened 2022-12-23
filter property in options object is not called for every copied file when recursive option is true


FreePhoenix888:
- **Operating System: Linux. Ubuntu 22.04**
- **Node.js version: 14.15.0**
- **`fs-extra` version: 11.1.0**

# Expected
`copySync` with `recursive: true` in options object calls `filter` with every file which is being copied

# Actual
`copySync` with `recursive: true` in options object calls `filter` with only src and dest directory paths

RyanZim:
Huh, I wonder when this broke. We should definitely have a unit test for this; if you've got a chance, a PR with a failing unit test would be greatly appreciated. Otherwise, I'll try to get to this soon.

FreePhoenix888:
@RyanZim , https://github.com/jprichardson/node-fs-extra/pull/989

RyanZim:
Closing as per https://github.com/jprichardson/node-fs-extra/pull/989#issuecomment-1365399194.

PR #989, opened 2022-12-24
Create copySync.test.js

FreePhoenix888:


FreePhoenix888:
https://github.com/jprichardson/node-fs-extra/issues/988

FreePhoenix888:
@jdalton , which of them are warnings and which of them are errors? https://github.com/jprichardson/node-fs-extra/actions/runs/3770641872/jobs/6413789692#step:5:13
I have fixed this problem: https://github.com/jprichardson/node-fs-extra/actions/runs/3770641872/jobs/6413789692#step:5:24

If other things are errors and must be processed - could not you please do it? You are allowed to edit this pull request:
image

RyanZim:
I finally got a chance to look at this and test in detail. fs-extra is working as intended here. First, the actual source and dest directory are passed to the filter. If the `filter` returns `false`, we stop copying, and make no further calls to `filter`. However, if it returns `true`, `filter` is called with each file in the directory.

PR #991, opened 2023-01-07
Fix build status badge

BendingBender:
Fix breaking change as described in https://github.com/badges/shields/issues/8671


RyanZim:
Thanks!

PR #994, opened 2023-02-02
Fix preserving timestamp when the move command is executing

KostiantynPopovych:
This PR resolves issue: #992.

These test files are pretty similar, I just took the `copy` method test files as a blueprint, and it seems we can make some refactoring here and there. I'm open to any suggestions!

PR #1001, opened 2023-03-06
Unify cross-device move test configuration/logic

RyanZim:
Especially since https://github.com/jprichardson/node-fs-extra/pull/994, functionally equivalent code was being needlessly duplicated. This centralizes into one place the configuration for locally testing cross-device moves; and unifies test code considerably.

### DIFF
--- a/lib/move/__tests__/cross-device-utils.js
+++ b/lib/move/__tests__/cross-device-utils.js
@@ -1,0 +1,21 @@
+const fs = require('graceful-fs')
+const path = require('path')
+
+// USER CONFIGURATION
+let runCrossDeviceTests = false // change to true to enable
+const differentDevice = '/mnt' // change this path to a mounted device
+
+if (runCrossDeviceTests) {
+  // make sure we have permission on device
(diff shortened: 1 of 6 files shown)


lib/move/__tests__/cross-device-utils.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('graceful-fs')
+const path = require('path')
+
+// USER CONFIGURATION
+let runCrossDeviceTests = false // change to true to enable
```
JPeer264, line 5: Wouldn't it make sense to use environment variables for these two settings? This would prevent commiting changes by accident.

A benefit would be that we could document that in the README and make it visible (with the downside of maintaining the README of these env variables).

RyanZim: Good idea; could actually unify to a single environment variable for the path; if it's set, we run the tests, if it's not set, we don't. Will refactor this when I get a chance.

JPeer264 (approved):

Issue #1020, opened 2023-10-19
[Proposal] Refactor the entire code base to Promise-based


SukkaW:
We can use `async` and `await` to organize the codebase and `universalify.fromPromise` to create Node.js-callback style APIs from the promise-based APIs.

RyanZim:
Yeah; that'd be great. It's a ton of work, but if you want to take a stab at it, you're more than welcome to! Just please make granular PRs; not one gigantic PR that takes hours to review.

SukkaW:
> Yeah; that'd be great. It's a ton of work, but if you want to take a stab at it, you're more than welcome to! Just please make granular PRs; not one gigantic PR that takes hours to review.

Sure! Let's do this one API migration per PR.

I will list the roadmap here for progress tracking:

- [ ] `copy` #1021
- [ ] `fs` (add `lib/fs/promise.js` to include the promisfied version of `graceful-fs` API)
- [ ] `empty`
- [ ] `ensure`
- [ ] `json`
- [ ] `mkdirs`
- [ ] `move`
- [ ] `output-file`
- [ ] `path-exists`
- [ ] `remove`

PR #1021, opened 2023-10-19
Refactor `copy` API in async/await

SukkaW:
The PR is the first part of proposal #1020:

- The PR only migrates the `copy` API. Other APIs will be migrated in the following PRs
- The PR adds Promise-based implementations for some utility functions. The callback-based implementations are retained, but they will be removed once the rest of the library has migrated to Promise-based implementations.

### DIFF
--- a/lib/copy/__tests__/copy-broken-symlink.test.js
+++ b/lib/copy/__tests__/copy-broken-symlink.test.js
@@ -5,7 +5,7 @@ const os = require('os')
 const fse = require('../..')
 const path = require('path')
 const assert = require('assert')
-const copy = require('../copy')
+const { copy } = require('../')
 
 /* global afterEach, beforeEach, describe, it */
 
--- a/lib/copy/copy.js
+++ b/lib/copy/copy.js
@@ -7,12 +7,14 @@ const { pathExists } = require('../path-exists')
 const { utimesMillisAsync } = require('../util/utimes')
 const stat = require('../util/stat')
 
-async function copy (src, dest, opts) {
+async function copy (src, dest, opts = {}) {
   if (typeof opts === 'function') {
     opts = { filter: opts }
   }
 
-  opts = opts || {}
+  opts = typeof opts === 'function'
+    ? { filter: opts }
+    : opts
 
   opts.clobber = 'clobber' in opts ? !!opts.clobber : true // default to true for now
   opts.overwrite = 'overwrite' in opts ? !!opts.overwrite : opts.clobber // overwrite falls back to clobber
@@ -42,12 +44,9 @@ async function checkParentDir (destStat, src, dest, opts) {
 
   const dirExists = await pathExists(destParent)
 
-  if (dirExists) return getStats(destStat, src, dest, opts)
-
-  const parentDirExists = await pathExists(destParent)
-  if (parentDirExists) return getStats(destStat, src, dest, opts)
-
-  await mkdirs(destParent)
+  if (!dirExists) {
+    await mkdirs(destParent)
+  }
 
   return getStats(destStat, src, dest, opts)
 }
@@ -75,12 +74,9 @@ async function getStats (destStat, src, dest, opts) {
   throw new Error(`Unknown file: ${src}`)
 }
 
-function onFile (srcStat, destStat, src, dest, opts) {
+async function onFile (srcStat, destStat, src, dest, opts) {
   if (!destStat) return copyFile(srcStat, src, dest, opts)
-  return mayCopyFile(srcStat, src, dest, opts)
-}
 
-async function mayCopyFile (srcStat, src, dest, opts) {
   if (opts.overwrite) {
     await fs.unlink(dest)
     return copyFile(srcStat, src, dest, opts)
@@ -104,7 +100,6 @@ async function handleTimestampsAndMode (srcMode, src, dest) {
   // (through utimes call)
   if (fileIsNotWritable(srcMode)) {
     await makeFileWritable(dest, srcMode)
-    return setDestTimestampsAndMode(srcMode, src, dest)
   }
   return setDestTimestampsAndMode(srcMode, src, dest)
 }
@@ -118,7 +113,12 @@ function makeFileWritable (dest, srcMode) {
 }
 
 async function setDestTimestampsAndMode (srcMode, src, dest) {
-  await setDestTimestamps(src, dest)
+  // The initial srcStat.atime cannot be trusted
+  // because it is modified by the read(2) system call
+  // (See https://nodejs.org/api/fs.html#fs_stat_time_values)
+  const updatedSrcStat = await fs.stat(src)
+  await utimesMillisAsync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
+
   return setDestMode(dest, srcMode)
 }
 
@@ -126,25 +126,14 @@ function setDestMode (dest, srcMode) {
   return fs.chmod(dest, srcMode)
 }
 
-async function setDestTimestamps (src, dest) {
-  // The initial srcStat.atime cannot be trusted
-  // because it is modified by the read(2) system call
-  // (See https://nodejs.org/api/fs.html#fs$stat_time_values)
-  const updatedSrcStat = await fs.stat(src)
-
-  return utimesMillisAsync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
-}
-
-function onDir (srcStat, destStat, src, dest, opts) {
-  if (!destStat) return mkDirAndCopy(srcStat.mode, src, dest, opts)
-  return copyDir(src, dest, opts)
-}
-
-async function mkDirAndCopy (srcMode, src, dest, opts) {
-  await fs.mkdir(dest)
+async function onDir (srcStat, destStat, src, dest, opts) {
+  if (!destStat) {
+    await fs.mkdir(dest)
+  }
   await copyDir(src, dest, opts)
-
-  return setDestMode(dest, srcMode)
+  if (!destStat) {
+    await setDestMode(dest, srcStat.mode)
+  }
 }
 
 async function copyDir (src, dest, opts) {
@@ -185,6 +174,9 @@ async function onLink (destStat, src, dest, opts) {
   try {
     resolvedDest = await fs.readlink(dest)
   } catch (e) {
+    // dest exists and is a regular file or directory,
+    // Windows may throw UNKNOWN error. If dest already exists,
+    // fs throws error anyway, so no need to guard against it here.
     if (e.code === 'EINVAL' || e.code === 'UNKNOWN') return fs.symlink(resolvedSrc, dest)
     throw e
   }
@@ -202,10 +194,6 @@ async function onLink (destStat, src, dest, opts) {
     throw new Error(`Cannot overwrite '${resolvedDest}' with '${resolvedSrc}'.`)
   }
 
-  return copyLink(resolvedSrc, dest)
-}
-
-async function copyLink (resolvedSrc, dest) {
   await fs.unlink(dest)
   return fs.symlink(resolvedSrc, dest)
 }
--- a/lib/copy/copy.js
+++ b/lib/copy/copy.js
@@ -2,20 +2,29 @@
 
 const fs = require('graceful-fs')
 const path = require('path')
-const mkdirs = require('../mkdirs').mkdirs
-const pathExists = require('../path-exists').pathExists
-const utimesMillis = require('../util/utimes').utimesMillis
+const { mkdirs } = require('../mkdirs')
+const { pathExists } = require('../path-exists')
+const { utimesMillisAsync } = require('../util/utimes')
 const stat = require('../util/stat')
 
-function copy (src, dest, opts, cb) {
-  if (typeof opts === 'function' && !cb) {
-    cb = opts
-    opts = {}
-  } else if (typeof opts === 'function') {
+const { promisify } = require('util')
+
+// TODO: re-exports promisified version of graceful-fs functions from lib/fs/promise.js
+const fs$stat = promisify(fs.stat)
+const fs$lstat = promisify(fs.lstat)
+const fs$unlink = promisify(fs.unlink)
+const fs$copyFile = promisify(fs.copyFile)
+const fs$mkdir = promisify(fs.mkdir)
+const fs$chmod = promisify(fs.chmod)
+const fs$readdir = promisify(fs.readdir)
+const fs$readlink = promisify(fs.readlink)
+const fs$symlink = promisify(fs.symlink)
+
+async function copy (src, dest, opts) {
+  if (typeof opts === 'function') {
     opts = { filter: opts }
   }
 
-  cb = cb || function () {}
   opts = opts || {}
 
   opts.clobber = 'clobber' in opts ? !!opts.clobber : true // default to true for now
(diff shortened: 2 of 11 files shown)


SukkaW:
@RyanZim The PR is ready to be reviewed now. `npm run lint` passes, and all test cases pass as well, check CI run here: https://github.com/SukkaW/node-fs-extra/actions/runs/6582727479

lib/copy/__tests__/copy-broken-symlink.test.js
```diff
@@ -5,7 +5,7 @@ const os = require('os')
 const fse = require('../..')
 const path = require('path')
 const assert = require('assert')
-const copy = require('../copy')
+const { copy } = require('../')
```
SukkaW, line 8: I need to change the import here. `lib/copy/copy.js` now exports the promise-only API (instead of the callback-only API), while `lib/copy/index.js` now exports the unified API.

lib/util/__tests__/utimes.test.js
```diff
@@ -33,7 +33,7 @@ describe('utimes', () => {
     fse.emptyDir(TEST_DIR, done)
     // reset stubs
     gracefulFsStub = {}
-    utimes = proxyquire('../utimes', { 'graceful-fs': gracefulFsStub })
+    utimes = proxyquire('../utimes', { '../fs': gracefulFsStub })
```
SukkaW, line 36: The mock here has to be replaced, as `lib/util/utimes.js` no longer uses `graceful-fs` directly. It now uses the unified API from `lib/fs/index.js`.

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
   // Make sure the file is writable before setting the timestamp
   // otherwise open fails with EPERM when invoked with 'r+'
   // (through utimes call)
   if (fileIsNotWritable(srcMode)) {
-    return makeFileWritable(dest, srcMode, err => {
-      if (err) return cb(err)
-      return setDestTimestampsAndMode(srcMode, src, dest, cb)
-    })
+    await makeFileWritable(dest, srcMode)
+    return setDestTimestampsAndMode(srcMode, src, dest)
```
RyanZim, line 107: No need to duplicate this inside the conditional since we're not inside a callback anymore; just allow it to fall through to the `setDestTimestampsAndMode` call below.

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
-function setDestTimestamps (src, dest, cb) {
+async function setDestTimestamps (src, dest) {
   // The initial srcStat.atime cannot be trusted
   // because it is modified by the read(2) system call
-  // (See https://nodejs.org/api/fs.html#fs_stat_time_values)
-  fs.stat(src, (err, updatedSrcStat) => {
-    if (err) return cb(err)
-    return utimesMillis(dest, updatedSrcStat.atime, updatedSrcStat.mtime, cb)
-  })
+  // (See https://nodejs.org/api/fs.html#fs$stat_time_values)
```
RyanZim, line 132: URL change here is accidental? New link is broken for me; old one still works.

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
+
+  return utimesMillisAsync(dest, updatedSrcStat.atime, updatedSrcStat.mtime)
 }
 
-function onDir (srcStat, destStat, src, dest, opts, cb) {
-  if (!destStat) return mkDirAndCopy(srcStat.mode, src, dest, opts, cb)
-  return copyDir(src, dest, opts, cb)
+function onDir (srcStat, destStat, src, dest, opts) {
+  if (!destStat) return mkDirAndCopy(srcStat.mode, src, dest, opts)
+  return copyDir(src, dest, opts)
```
RyanZim, line 140: Can reduce code duplication by inlining `mkDirAndCopy` like this:
```suggestion
  if (!destStat) await fs.mkdir(dest)
  await copyDir(src, dest, opts)
  if (!destStat) await setDestMode(dest, srcMode)
```

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
+function copyDirItems (items, src, dest, opts) {
   const item = items.pop()
-  if (!item) return cb()
-  return copyDirItem(items, item, src, dest, opts, cb)
+  if (!item) return
+  return copyDirItem(items, item, src, dest, opts)
 }
 
-function copyDirItem (items, item, src, dest, opts, cb) {
+async function copyDirItem (items, item, src, dest, opts) {
```
RyanZim, line 161: This logic needs to be rethought/reworked. We currently use a recursive `copyDirItems` call, which mutates `items`. This made sense to avoid the complexity of serially iterating a callback-based process, but is non-idiomatic in an async/await world. I'm thinking we use a `for` loop in `copyDir` to iterate over the items...I'll leave the exact details to you.

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
+  }
+  if (!destStat) {
+    return fs.symlink(resolvedSrc, dest)
+  }
+
+  let resolvedDest = null
+  try {
+    resolvedDest = await fs.readlink(dest)
+  } catch (e) {
+    if (e.code === 'EINVAL' || e.code === 'UNKNOWN') return fs.symlink(resolvedSrc, dest)
```
RyanZim, line 188: Please preserve explanatory comments from original code

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
+    throw new Error(`Cannot overwrite '${resolvedDest}' with '${resolvedSrc}'.`)
+  }
+
+  return copyLink(resolvedSrc, dest)
+}
+
+async function copyLink (resolvedSrc, dest) {
+  await fs.unlink(dest)
+  return fs.symlink(resolvedSrc, dest)
 }
```
RyanZim, line 211: Can inline this logic:
```suggestion
  await fs.unlink(dest)
  return fs.symlink(resolvedSrc, dest)
}
```

RyanZim (changes requested):
I've left a several code comments. Additionally, I have a few other concerns:

1. `copy` is not expected to return a value, yet we use `return` throughout the code for final async calls inside functions; this could result in unintentionally returning a value in the future; perhaps we should use `await` instead?
2. Is there a reason for duplicating the util functions vs. just wrapping them in universalify? We could wrap them in `fromCallback` here, then port them to promise implementations and wrap in `fromPromise` in a separate PR. Anything I'm missing? (I have not reviewed those changes yet)

lib/copy/copy.js
```diff
@@ -1,21 +1,17 @@
...
-    cb = opts
-    opts = {}
-  } else if (typeof opts === 'function') {
+async function copy (src, dest, opts) {
+  if (typeof opts === 'function') {
     opts = { filter: opts }
   }
 
-  cb = cb || function () {}
   opts = opts || {}
```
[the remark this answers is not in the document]
SukkaW: Done in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

I changed to use the default assign syntax (no nested condition!):

```js
// If `opts` is not specified or undefined, assign {}
async function copy (src, dest, opts = {}) {
  // now re-assign `opts` if it is a function
  if (typeof opts === 'function') {
    opts = { filter: opts }
  }
```

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
-      return getStats(destStat, src, dest, opts, cb)
-    })
-  })
+
+  const dirExists = await pathExists(destParent)
+
+  if (dirExists) return getStats(destStat, src, dest, opts)
+
+  const parentDirExists = await pathExists(destParent)
+  if (parentDirExists) return getStats(destStat, src, dest, opts)
```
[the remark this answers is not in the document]
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

lib/copy/copy.js
```diff
@@ -30,209 +26,188 @@ function copy (src, dest, opts, cb) {
...
+  if (srcStat.isFIFO()) throw new Error(`Cannot copy a FIFO pipe: ${src}`)
+  throw new Error(`Unknown file: ${src}`)
 }
 
-function onFile (srcStat, destStat, src, dest, opts, cb) {
-  if (!destStat) return copyFile(srcStat, src, dest, opts, cb)
-  return mayCopyFile(srcStat, src, dest, opts, cb)
+function onFile (srcStat, destStat, src, dest, opts) {
+  if (!destStat) return copyFile(srcStat, src, dest, opts)
+  return mayCopyFile(srcStat, src, dest, opts)
```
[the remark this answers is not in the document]
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c). The `mayCopyFile` is now inlined.

lib/copy/copy.js
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

lib/copy/copy.js
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

lib/copy/copy.js
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

lib/copy/copy.js
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

lib/copy/copy.js
SukkaW: Fixed in [`bc8ee0b` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/bc8ee0bf340f04fad0fed408bbdd0cc3747e9d0c).

And the entire `copyLink` has been inlined into `onLink` as well.

lib/copy/copy.js
SukkaW: Refactored this part in [`cf3aed7` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/cf3aed7a1d413e329fa7edede323f2479f3901a0).

The `onDir` now uses a for loop to iterate every item inside the directory (from `await fs.readdir`), perform the pre-copy check (the `filter` option and the `stat.checkPathsAsync`), and then invoke the `getStatsAndPerformCopy` function (which would recursively call `onDir` again).

lib/util/__tests__/utimes.test.js
```diff
@@ -68,25 +70,25 @@ describe('utimes', () => {
     it('should close open file desciptors after encountering an error', done => {
       const fakeFd = Math.random()
 
-      gracefulFsStub.open = (pathIgnored, flagsIgnored, modeIgnored, callback) => {
+      gracefulFsStub.open = u((pathIgnored, flagsIgnored, modeIgnored, callback) => {
```
SukkaW, line 73: The `gracefulFsStub` needs to be universalify as the `utimesMillis` function is now universalised.

SukkaW:
@RyanZim I've addressed all the code review suggestions. I've also universalized the related utility functions to reduce the duplicated implementations. The CI still passes after the latest changes: https://github.com/SukkaW/node-fs-extra/actions/runs/6598480115

Note that the `coprDir` has been heavily refactored (no more `items` array mutation!).

lib/copy/copy.js
```diff
@@ -122,43 +108,47 @@ async function setDestTimestampsAndMode (srcMode, src, dest) {
...
-}
 
-function copyDirItems (items, src, dest, opts) {
-  const item = items.pop()
-  if (!item) return
-  return copyDirItem(items, item, src, dest, opts)
-}
+  // loop through the files in the current directory to copy everything
+  for (let i = 0, len = items.length; i < len; i++) {
+    const item = items[i]
```
RyanZim, line 120: Why not use a `for..of` loop here? Sorry if my initial review suggested I wanted a plain `for` loop.

lib/util/utimes.js
```diff
@@ -3,7 +3,7 @@
 const fs = require('../fs')
 const u = require('universalify').fromPromise
 
-async function utimesMillis (path, atime, mtime) {
+async function utimesMillis (path, atime, mtime, cb) {
```
RyanZim, line 6: Why `cb` here?

lib/copy/copy.js
```diff
@@ -30,209 +24,156 @@ function copy (src, dest, opts, cb) {
...
 }
 
-function setDestTimestampsAndMode (srcMode, src, dest, cb) {
-  setDestTimestamps(src, dest, err => {
-    if (err) return cb(err)
-    return setDestMode(dest, srcMode, cb)
-  })
-}
+async function onDir (srcStat, destStat, src, dest, opts) {
+  // the dest direcotry might not exist, create it
```
RyanZim, line 111:
```suggestion
  // the dest directory might not exist, create it
```

lib/util/utimes.js
```diff
@@ -1,17 +1,33 @@
...
+
+  try {
+    await fs.close(fd)
+  } catch (e) {
+    closeErr = e
+  }
+
+  if (futimesErr) {
+    throw futimesErr
+  }
```
RyanZim, line 27: We can achieve the same behavior a little more simply with `try..finally`.
```suggestion
  let closeErr
  try {
    await fs.futimes(fd, atime, mtime)
  } finally {
    // put this in a try..catch so the error from futimes is thrown preferentially to any error here
    try {
      await fs.close(fd)
    } catch (e) {
      closeErr = e
    }
  }
```

RyanZim (changes requested):
Just a couple comments

lib/util/utimes.js
SukkaW: Removed in [`1ce5d34` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/1ce5d348c426ea20c5006389477e163211898b22).

lib/copy/copy.js
SukkaW: Changed in [`1ce5d34` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/1ce5d348c426ea20c5006389477e163211898b22).

lib/copy/copy.js
SukkaW: Typo fixed in [`1ce5d34` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/1ce5d348c426ea20c5006389477e163211898b22).

lib/util/utimes.js
SukkaW: Changed in [`1ce5d34` (#1021)](https://github.com/jprichardson/node-fs-extra/pull/1021/commits/1ce5d348c426ea20c5006389477e163211898b22).

RyanZim (approved):

PR #1023, opened 2023-10-21
Refactor `create*`/`ensure*` API in async/await

SukkaW:
The PR is a part of the proposal https://github.com/jprichardson/node-fs-extra/issues/1020.

All test cases pass: https://github.com/SukkaW/node-fs-extra/actions/runs/6599300023.

### DIFF
--- a/lib/ensure/file.js
+++ b/lib/ensure/file.js
@@ -5,7 +5,7 @@ const path = require('path')
 const fs = require('../fs')
 const mkdir = require('../mkdirs')
 
-async function createFile (file) {
+async function createFile(file) {
   let stats
   try {
     stats = await fs.stat(file)
@@ -28,18 +28,16 @@ async function createFile (file) {
     }
   }
 
-  if (dirStats) {
-    if (dirStats.isDirectory()) {
-      await fs.writeFile(file, '')
-    } else {
-      // parent is not a directory
-      // This is just to cause an internal ENOTDIR error to be thrown
-      await fs.readdir(dir)
-    }
+  if (dirStats.isDirectory()) {
+    await fs.writeFile(file, '')
+  } else {
+    // parent is not a directory
+    // This is just to cause an internal ENOTDIR error to be thrown
+    await fs.readdir(dir)
   }
 }
 
-function createFileSync (file) {
+function createFileSync(file) {
   let stats
   try {
     stats = fs.statSync(file)
--- a/lib/ensure/file.js
+++ b/lib/ensure/file.js
@@ -6,8 +6,11 @@ const fs = require('../fs')
 const mkdir = require('../mkdirs')
 
 async function createFile (file) {
-  const fileStats = await fs.stat(file)
-  if (fileStats.isFile()) return
+  let stats
+  try {
+    stats = await fs.stat(file)
+  } catch { }
+  if (stats && stats.isFile()) return
 
   const dir = path.dirname(file)
 
--- a/lib/ensure/file.js
+++ b/lib/ensure/file.js
@@ -5,7 +5,7 @@ const path = require('path')
 const fs = require('../fs')
 const mkdir = require('../mkdirs')
 
-async function createFile(file) {
+async function createFile (file) {
   let stats
   try {
     stats = await fs.stat(file)
@@ -37,7 +37,7 @@ async function createFile(file) {
   }
 }
 
-function createFileSync(file) {
+function createFileSync (file) {
   let stats
   try {
     stats = fs.statSync(file)
--- a/lib/ensure/file.js
+++ b/lib/ensure/file.js
@@ -1,50 +1,46 @@
 'use strict'
 
-const u = require('universalify').fromCallback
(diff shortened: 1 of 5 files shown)


lib/ensure/file.js
```diff
@@ -1,50 +1,49 @@
...
-      else {
-        // parent is not a directory
-        // This is just to cause an internal ENOTDIR error to be thrown
-        fs.readdir(dir, err => {
-          if (err) return callback(err)
-        })
-      }
-    })
-  })
+  if (dirStats) {
```
RyanZim, line 31: This `if` block isn't needed, since we invariably `return` or `throw` in the `catch` block above

lib/ensure/link.js
```diff
@@ -1,39 +1,39 @@
...
-      if (dstStat && areIdentical(srcStat, dstStat)) return callback(null)
+  let srcStat
+  try {
+    srcStat = await fs.lstat(srcpath)
+  } catch (err) {
+    err.message = err.message.replace('lstat', 'ensureLink')
+    throw err
+  }
+
+  if (srcStat && dstStat && areIdentical(srcStat, dstStat)) return
```
RyanZim, line 26: Again, we're throwing in the catch block, so no need to check for the existence of the variable:
```suggestion
  if (dstStat && areIdentical(srcStat, dstStat)) return
```

lib/ensure/symlink.js
```diff
@@ -1,65 +1,50 @@
...
-        if (dirExists) return fs.symlink(srcpath, dstpath, type, callback)
-        mkdirs(dir, err => {
-          if (err) return callback(err)
-          fs.symlink(srcpath, dstpath, type, callback)
-        })
-      })
-    })
-  })
+  await mkdirs(dir)
+  return fs.symlink(srcpath, dstpath, toType)
```
RyanZim, line 40: Could reduce with `if (!dirExists) await mkdirs(dir)`

lib/ensure/symlink.js
```diff
@@ -1,65 +1,50 @@
...
 
-const pathExists = require('../path-exists').pathExists
+const { pathExists } = require('../path-exists')
 
 const { areIdentical } = require('../util/stat')
 
-function createSymlink (srcpath, dstpath, type, callback) {
-  callback = (typeof type === 'function') ? type : callback
-  type = (typeof type === 'function') ? false : type
+async function createSymlink (srcpath, dstpath, type = false) {
```
RyanZim, line 16: Not sure why the old logic worked this way, but no reason to default to `false` here.
```suggestion
async function createSymlink (srcpath, dstpath, type) {
```

lib/ensure/symlink-type.js
```diff
@@ -1,22 +1,25 @@
...
-  callback = (typeof type === 'function') ? type : callback
-  type = (typeof type === 'function') ? false : type
-  if (type) return callback(null, type)
-  fs.lstat(srcpath, (err, stats) => {
-    if (err) return callback(null, 'file')
-    type = (stats && stats.isDirectory()) ? 'dir' : 'file'
-    callback(null, type)
-  })
-}
+async function symlinkType (srcpath, type = false) {
```
RyanZim, line 6: Not sure why the old logic worked this way, but no reason to default to `false` here.
```suggestion
async function symlinkType (srcpath, type) {
```

RyanZim (changes requested):

lib/ensure/file.js
SukkaW: Fixed in [`fbfc9c7` (#1023)](https://github.com/jprichardson/node-fs-extra/pull/1023/commits/fbfc9c7d86925105cf905df9e11b51132be386d4).

lib/ensure/link.js
SukkaW: Fixed in [`fbfc9c7` (#1023)](https://github.com/jprichardson/node-fs-extra/pull/1023/commits/fbfc9c7d86925105cf905df9e11b51132be386d4).

lib/ensure/symlink.js
SukkaW: Fixed in [`fbfc9c7` (#1023)](https://github.com/jprichardson/node-fs-extra/pull/1023/commits/fbfc9c7d86925105cf905df9e11b51132be386d4).

lib/ensure/symlink-type.js
SukkaW: Fixed in [`fbfc9c7` (#1023)](https://github.com/jprichardson/node-fs-extra/pull/1023/commits/fbfc9c7d86925105cf905df9e11b51132be386d4).

lib/ensure/symlink.js
SukkaW: Fixed in [`fbfc9c7` (#1023)](https://github.com/jprichardson/node-fs-extra/pull/1023/commits/fbfc9c7d86925105cf905df9e11b51132be386d4).

RyanZim (approved):

PR #1024, opened 2023-10-21
Refactor `outputFile` APi in async/await

SukkaW:
The PR is a part of the proposal https://github.com/jprichardson/node-fs-extra/issues/1020.

All test cases pass: https://github.com/SukkaW/node-fs-extra/actions/runs/6599334843.

### DIFF
--- a/lib/output-file/index.js
+++ b/lib/output-file/index.js
@@ -1,36 +1,29 @@
 'use strict'
 
-const u = require('universalify').fromCallback
-const fs = require('graceful-fs')
+const u = require('universalify').fromPromise
+const fs = require('../fs')
 const path = require('path')
 const mkdir = require('../mkdirs')
 const pathExists = require('../path-exists').pathExists
 
-function outputFile (file, data, encoding, callback) {
-  if (typeof encoding === 'function') {
-    callback = encoding
-    encoding = 'utf8'
+async function outputFile (file, data, encoding = 'utf-8') {
+  const dir = path.dirname(file)
+
+  const itDoes = await pathExists(dir)
+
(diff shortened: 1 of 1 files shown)


lib/output-file/index.js
```diff
@@ -1,36 +1,29 @@
...
-
+async function outputFile (file, data, encoding = 'utf-8') {
   const dir = path.dirname(file)
-  pathExists(dir, (err, itDoes) => {
-    if (err) return callback(err)
-    if (itDoes) return fs.writeFile(file, data, encoding, callback)
 
-    mkdir.mkdirs(dir, err => {
-      if (err) return callback(err)
+  const itDoes = await pathExists(dir)
```
RyanZim, line 12: Let's rename this to something sensible like `dirExists`, while we're here :grinning:

lib/output-file/index.js
```diff
@@ -1,36 +1,29 @@
...
 
-      fs.writeFile(file, data, encoding, callback)
-    })
-  })
+  if (!itDoes) {
```
RyanZim, line 14: Alternately, just inline `await pathExists(dir)` in the `if` expression, to match the sync version.

RyanZim (commented):
LGTM, just one stupid variable name from the old code

lib/output-file/index.js
SukkaW: Fixed in [`990ffd2` (#1024)](https://github.com/jprichardson/node-fs-extra/pull/1024/commits/990ffd25209b8e6b4e9dd2a8c07d387cbaafbb53)

RyanZim (approved):

PR #1025, opened 2023-10-25
Refactor `move` API to async/await

SukkaW:
The PR is the final part of #1020.

### DIFF
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -6,7 +6,6 @@ const fse = require('../../')
 const path = require('path')
 const assert = require('assert')
 const { differentDevice, ifCrossDeviceEnabled } = require('./cross-device-utils')
-const u = require('universalify').fromPromise
 
 /* global afterEach, beforeEach, describe, it */
 
--- a/lib/move/__tests__/move.test.js
+++ b/lib/move/__tests__/move.test.js
@@ -1,25 +1,24 @@
 'use strict'
 
-const fs = require('graceful-fs')
+const fs = require('../../fs')
 const os = require('os')
 const fse = require('../../')
 const path = require('path')
 const assert = require('assert')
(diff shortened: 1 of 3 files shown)


lib/move/__tests__/move.test.js
```diff
@@ -12,14 +12,12 @@ const { differentDevice, ifCrossDeviceEnabled } = require('./cross-device-utils'
...
-    setTimeout(() => {
-      const err = new Error()
-      err.code = errCode
-      callback(err)
-    }, 10)
+    const err = new Error()
+    err.code = errCode
+
+    return Promise.reject(err)
   }
```
SukkaW, line 21: The `move` now uses the Promise version of the `fs.rename` API (instead of the callback version), so the stub here needs to be changed.

lib/move/move.js
```diff
@@ -1,76 +1,59 @@
...
-        return doRename(src, dest, overwrite, isChangingCase, cb)
-      })
-    })
-  })
-}
+  // If the parent of dest is not root, make sure it exists before proceeding
+  const destParent = path.dirname(dest)
+  const parsedParentPath = path.parse(destParent)
+  if (parsedParentPath.root !== destParent) {
+    await mkdirp(path.dirname(dest))
```
RyanZim, line 22: We already have the dirname, let's use it
```suggestion
    await mkdirp(destParent)
```

RyanZim (changes requested):
One nit; otherwise LGTM

lib/move/move.js
SukkaW: Fixed in [`7ab00d5` (#1025)](https://github.com/jprichardson/node-fs-extra/pull/1025/commits/7ab00d58bf1f018aaa2f3490e52c5fa401c7c449).

JPeer264 (approved):
Amazing work!

RyanZim (approved):

RyanZim:
Thanks a lot @SukkaW

PR #1026, opened 2023-10-27
Perf: copying dir contents in parallel

SukkaW:
The PR uses a technique similar to #885 to speed up the recursive copy operation.

As `fs-extra` is built on top of `graceful-fs`, this change won't cause `EMFILE: too many open files` as `graceful-fs` will exponentially backoff and retry.

### DIFF
--- a/lib/copy/copy.js
+++ b/lib/copy/copy.js
@@ -113,21 +113,23 @@ async function onDir (srcStat, destStat, src, dest, opts) {
     await fs.mkdir(dest)
   }
 
+  const items = await fs.readdir(src)
+
   // loop through the files in the current directory to copy everything
-  for (const item of await fs.readdir(src)) {
+  await Promise.all(items.map(async item => {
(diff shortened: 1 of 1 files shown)


RyanZim (approved):
Change looks good to me; but would like a second pair of eyes here to make sure I'm not missing something.

RyanZim:
@manidlou bump

manidlou (approved):
LGTM

RyanZim:
Published in v11.2.0!

SukkaW:
@RyanZim ❤️ 

Too bad I can't port this change to Node.js' `fs#cp`. Node.js doesn't have `graceful-fs` supports, thus `EMFILE`can not be avoided.

RyanZim:
> Too bad I can't port this change to Node.js' fs#cp. Node.js doesn't have graceful-fs, thus EMFILEcan not be avoided.

You'd have to implement a concurrency limit yourself, I guess.

PR #1036, opened 2024-03-03
Update copyright in README.md

ISlavaRa:


RyanZim:
If we're going to update this; we should update the LICENSE file as well

ISlavaRa:
> If we're going to update this; we should update the LICENSE file as well

Done ;)

Wadey9111 (approved):

PR #1045, opened 2025-01-10
Add promise support for missing fs methods

### LINKED ISSUE #1044: Support for statfs() and other newer fs functions in more recent versions of node.js
Are newer fs functions "automatically" included in the promisified list of functions, or must these functions specifically be listed in the fs-extra source?

RyanZim:
Fixes #1044

Somehow, I added a comment about `fs.cp` but never added it to the list :man_shrugging:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         node: [14.x, 16.x, 18.x, 19.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-13, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


RyanZim:
CI is failing because `macos-latest` now runs on arm64, but old Node versions only support x64 arch

RyanZim:
Temporarily fixed the macOS version to work around this

JPeer264:
> CI is failing because `macos-latest` now runs on arm64, but old Node versions only support x64 arch

For that it would make sense to drop the support for 14.x and 16.x, and start supporting newer versions, at least in GitHub Actions. WDYT?

JPeer264 (approved):

RyanZim:
Yeah, we should drop old Node versions, but that'd be a separate breaking change, and I'd like to do some thinking about other breaking changes we should include in the major release.

PR #1047, opened 2025-04-09
Use `import` conditional exports for default export for esm modules.

ycmjason:
Hello!

I'd like to propose using the `import` [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) to export `/esm`.

Thank you for your consideration!

Happy to update the doc if this is something we'd like to move forward! Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,10 @@
   },
   "main": "./lib/index.js",
   "exports": {
-    ".": "./lib/index.js",
+    ".": {
+      "import": "./lib/esm.mjs",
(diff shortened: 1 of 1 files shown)


JPeer264:
It was actively decided against it back in the days https://github.com/jprichardson/node-fs-extra/pull/974#discussion_r1027124236. I assume this comment still remains valid @RyanZim ?

ycmjason:
Very true and fair. Thank you for pointing out! Closing the PR now!

PR #1052, opened 2025-08-01
Do not mutate args in ensure symlink tests

RyanZim:
Just a bad practice period, and actually results in a callback being passed to the synchronous functions later in the file that reuse the same `args` array.


manidlou (approved):
LGTM, thank you @RyanZim

JPeer264 (approved):

PR #1065, opened 2026-04-09
Fix error handling in utimesMillis

RyanZim:
This was broken in https://github.com/jprichardson/node-fs-extra/pull/1021

Also remove old comment


manidlou (approved):
LGTM, thanks @RyanZim

RyanZim:
Just realized this isn't quite correct; if both `futimes` and `close` error, the close error overwrites the earlier error. Will fix after https://github.com/jprichardson/node-fs-extra/pull/1066 is merged.

PR #1068, opened 2026-04-20
Ensure all usages of areIdentical receive bigint stats

RyanZim:
On Windows, `stats.ino` can exceed `MAX_SAFE_INTEGER`.

We use `{ bigint: true }` here:
https://github.com/jprichardson/node-fs-extra/blob/1e9c57de9fc5e766d3c9938fc7289080c444a568/lib/util/stat.js#L8-L10

But there are a few other uses of `areIdentical()` that didn't get bigint stats. This fixes that.


JPeer264 (approved):
LGTM.

PR #1073, opened 2026-06-15
Fix copy/move recursing into source through a symlinked destination ancestor

### LINKED ISSUE #1071: copy/copySync can recurse into the source tree through a symlinked destination ancestor
I'd like to open a security advisory for this vulnerability we found. 

fs-extra's copy self-protection can be bypassed when an existing ancestor of the destination path is a symlink into the source tree and a deeper destination parent does not yet exist. The pre-copy ancestry check returns early on ENOENT, the missing path is then created under the symlink target, and recursive copy walks into the subtree it is creating until path growth or resource exhaustion stops it.

sarathfrancis90:
Fixes #1071.

`checkParentPaths` walks up the destination's ancestors to catch the case where the destination is actually inside the source, which otherwise makes `copy` recurse into the very tree it is creating. When an intermediate destination parent doesn't exist yet, it returned early on `ENOENT` and stopped checking. If a *deeper* ancestor is a symlink into the source tree, that early return bypasses the check entirely.

The result is that `copy(src, link/sub/dest)` (where `link` -> `src` and `link/sub` doesn't exist yet) keeps copying `src` into itself, growing `src/sub/dest/sub/dest/...` until it dies with `ENAMETOOLONG`, instead of raising the usual "Cannot copy to a subdirectory of itself" error.

Per @RyanZim's suggestion in the issue, the fix recurses to the next parent on `ENOENT` instead of returning, so the check keeps walking up until it reaches an existing ancestor (or the source parent / root). This affects `copy`, `copySync`, `move`, and `moveSync` since they all go through `checkParentPaths`.

### Testing
- Added regression tests for `copy` and `copySync` covering a non-existent dest parent under a symlink-into-src ancestor (fail before this change: async times out from the runaway recursion, sync doesn't throw; both pass after).
- Verified legitimate deep copies into multiple non-existent parents still succeed (no false positives).
- Full suite green: `npm test` (lint + 728 unit + ESM).

### DIFF
--- a/lib/copy/__tests__/copy-prevent-copying-into-itself.test.js
+++ b/lib/copy/__tests__/copy-prevent-copying-into-itself.test.js
@@ -210,6 +210,29 @@ describe('+ copy() - prevent copying into itself', () => {
         })
       })
 
+      it('should error when dest is a subdirectory of src and the dest parent does not exist yet', done => {
+        const destLink = path.join(TEST_DIR, 'dest-symlink')
+        fs.symlinkSync(src, destLink, 'dir')
+
+        const srclenBefore = klawSync(src).length
+        assert(srclenBefore > 2)
+
+        // The immediate dest parent ('sub') does not exist yet, but a deeper
+        // ancestor ('dest-symlink') is a symlink into src.
+        const dest = path.join(destLink, 'sub', 'dest')
+        assert(!fs.existsSync(path.dirname(dest)))
+        fs.copy(src, dest, err => {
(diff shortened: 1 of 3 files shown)


JPeer264 (approved):
Implementation and tests look fine, but I'm not the best to judge.

JPeer264:
Also thanks a lot for taking your time to fix it

RyanZim (approved):

RyanZim:
Published in `fs-extra@11.3.6`; sorry for the delay.

PR #1076, opened 2026-06-30
Throw EEXIST when ensureSymlink dest is a broken symlink

### LINKED ISSUE #925: Better error message for ensureSymlink if the current existing link is broken
- **Operating System:** Pop!_OS 20.04 LTS
- **Node.js version:** v16.11.0
- **`fs-extra` version:** 10.0.0

I'm not sure if this is the expected behavior.

Here's what I'm trying to do... basically I want to replicate what `ln -sfn` does. I would like to create a link to a file in my home bin folder. If I move the target file to another directory, then as expected, that link will now be broken. That's why I want to recreate that link, updating the target path. The problem is when I do this (calling `createSymlink`) node throws an error (because the link is broken):

```sh
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[Error: ENOENT: no such file or directory, stat '/home/myusername/bin/rgb-toggle'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/home/myusername/bin/rgb-toggle'
}
```

The error is thrown by Node when calling `fs.stat` or `fs.statSync`, here's what happens when I call it in Node REPL:

```js
> fs.statSync(`${os.homedir()}/bin/rgb-toggle`)
Uncaught Error: ENOENT: no such file or directory, stat '/home/myusername/bin/rgb-toggle'
    at Object.statSync (node:fs:1536:3) {
  errno: -2,
  syscall: 'stat',
  code: 'ENOENT',
  path: '/home/myusername/bin/rgb-toggle'
}
```

The link exists, only it's broken. Looks like that is expected behavior for Node because it will call POSIX `stat`, and as the [description][1]:

> stat() retrieve information about the file pointed to by pathname.

However for that purpose we can use `fs.lstat` or `fs.lstatSync` which will call POSIX `lstat`, which as [described][1]:

> lstat() is identical to stat(), except that if pathname is a symbolic link, then it returns information about the link itself, not the file that the link refers to.

Calling `fs.lstatSync` succeeds:

```js
> fs.lstatSync(`${os.homedir()}/bin/rgb-toggle`)
Stats {
  dev: 66307,
  mode: 41471,
  nlink: 1,
  uid: 1000,
  gid: 1000,
  rdev: 0,
  blksize: 4096,
  ino: 5396104,
  size: 60,
  blocks: 8,
  atimeMs: 1634260915813.8984,
  mtimeMs: 1634260915813.8984,
  ctimeMs: 1634260915813.8984,
  birthtimeMs: 1634260915813.8984,
  atime: 2021-10-15T01:21:55.814Z,
  mtime: 2021-10-15T01:21:55.814Z,
  ctime: 2021-10-15T01:21:55.814Z,
  birthtime: 2021-10-15T01:21:55.814Z
}
```

Shouldn't `fs.ensureSymlink` check for broken symlink and update it (given the new target path exists)?

[1]: https://man7.org/linux/man-pages/man2/lstat.2.html

sarathfrancis90:
Fixes #925.

`ensureSymlink`/`ensureSymlinkSync` stat the destination to check whether an existing symlink already points at `srcpath`. When that symlink is broken, following it fails and the raw `ENOENT` (`stat '<dest>'`) bubbles up to the caller. That's surprising, since a symlink pointing at a *different* existing target already throws `EEXIST` — only the broken case leaked the stat error. @RyanZim noted the same direction in the issue.

I skip the idempotency check when the destination can't be followed, so the existing creation path reports `EEXIST` consistently whether the existing link is broken or just points elsewhere. Added async and sync regression tests; `npm test` passes locally.

### DIFF
--- a/lib/ensure/symlink.js
+++ b/lib/ensure/symlink.js
@@ -35,8 +35,13 @@ async function createSymlink (srcpath, dstpath, type) {
       }
     }
 
-    const dstStat = await fs.stat(dstpath, { bigint: true })
-    if (areIdentical(srcStat, dstStat)) return
+    // Following the existing symlink fails if it's broken; in that case fall
+    // through so fs.symlink reports EEXIST, rather than leaking the stat ENOENT.
+    let dstStat
+    try {
+      dstStat = await fs.stat(dstpath, { bigint: true })
+    } catch { }
+    if (dstStat && areIdentical(srcStat, dstStat)) return
   }
 
   const relative = await symlinkPaths(srcpath, dstpath)
@@ -72,8 +77,13 @@ function createSymlinkSync (srcpath, dstpath, type) {
       }
     }
 
-    const dstStat = fs.statSync(dstpath, { bigint: true })
-    if (areIdentical(srcStat, dstStat)) return
+    // Following the existing symlink fails if it's broken; in that case fall
+    // through so fs.symlinkSync reports EEXIST, rather than leaking the stat ENOENT.
+    let dstStat
+    try {
+      dstStat = fs.statSync(dstpath, { bigint: true })
+    } catch { }
+    if (dstStat && areIdentical(srcStat, dstStat)) return
   }
 
   const relative = symlinkPathsSync(srcpath, dstpath)
--- a/lib/ensure/__tests__/symlink.test.js
+++ b/lib/ensure/__tests__/symlink.test.js
@@ -399,4 +399,27 @@ describe('fse-ensure-symlink', () => {
       assert.strictEqual(content, 'content')
     })
   })
+
+  // https://github.com/jprichardson/node-fs-extra/issues/925
+  describe('ensureSymlink() when dest is an existing broken symlink (issue #925)', () => {
+    it('should fail with EEXIST (not ENOENT) when dest is a broken symlink', async () => {
+      const srcpath = path.join(TEST_DIR, 'src-925.txt')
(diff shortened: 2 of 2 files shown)


lib/ensure/symlink.js
```diff
@@ -35,8 +35,13 @@ async function createSymlink (srcpath, dstpath, type) {
...
     }
 
-    const dstStat = await fs.stat(dstpath, { bigint: true })
-    if (areIdentical(srcStat, dstStat)) return
+    // Following the existing symlink fails if it's broken; in that case fall
+    // through so fs.symlink reports EEXIST, rather than leaking the stat ENOENT.
+    let dstStat
+    try {
+      dstStat = await fs.stat(dstpath, { bigint: true })
+    } catch { }
```
RyanZim: Seems like it might be wise to ensure the error is `ENOENT` before swallowing, and throw otherwise.

sarathfrancis90:
Good call. I've narrowed that catch so it only swallows ENOENT (the broken-link case we're actually expecting) and rethrows anything else, following the `err.code` pattern used elsewhere in the lib. Added tests for both the async and sync paths to make sure a non-ENOENT error now propagates.

RyanZim (approved):

RyanZim:
Published in `fs-extra@11.4.0` :tada:

PR #1080, opened 2026-09-10
fix: propagate emptyDir directory read errors

xianjianlf2:
`emptyDir()` and `emptyDirSync()` currently treat every `readdir` failure as a missing directory. If the directory exists but cannot be read, recursive `mkdir` can succeed and the operation reports success without clearing anything.

Only create the directory after `ENOENT`; propagate other read errors unchanged. Add portable regression tests for `EACCES` and `EIO` in the Promise, callback, and synchronous interfaces. All five new tests fail before the fix; existing missing-directory tests still pass.

Also reproduced on a real macOS directory with mode `000`: the original implementation returns success with its file untouched, while the patch reports `EACCES`.

Validation: `npm test` passes (Standard lint, 737 unit tests with 8 existing skips, and ESM tests). The changed empty-directory module has 100% coverage.

AI assistance: implemented and tested with OpenAI Codex.


RyanZim (approved):
Good catch, and thanks for submitting the PR!

One thing this will change is that if the specified path is a file, it will now error with `ENOTDIR` instead of `EEXIST`. This does make more sense, but I'm unsure if anyone is depending on this specific error code (i.e. should this be considered a breaking change?).

RyanZim:
Published in v11.4.1 :tada:

PR #1081, opened 2026-09-17
Add tests for capitalization change at the same time

RyanZim:


RyanZim:
Changes pushed to https://github.com/jprichardson/node-fs-extra/pull/1079

Issue #912, opened 2021-07-15
Naming conflict with Node.js core


RyanZim:
As per https://github.com/nodejs/node/pull/39372; Node.js is considering moving `copy*()` to `fs` core. This will create a naming conflict between our method and `fs` core when it's released. How do we want to handle this?

bcoe:
we've since moved to `fs.cp`.

RyanZim:
Closing as this won't be an issue with the current direction.

manidlou:
I figured I could reopen this instead of creating a new issue :grin: 

So another thing that we need to think about here at `fs-extra` is the options for copy method because considering latest changes on [nodejs/node#39372](https://github.com/nodejs/node/pull/39372) these are some differences between node core `fs.cp()` and `fse.copy()` options:

- There is a `recursive` option in `fs.cp()` that is default to `false`. In `fs-extra` we copy the directory recursively by default.
- The `overwrite` option is renamed to `force` in `fs.cp()`.

When [nodejs/node#39372](https://github.com/nodejs/node/pull/39372) gets landed in node core, we probably want to use it wherever is possible but as long as we support older node versions (without `fs.cp()`) we would fallback to our current (soon to be fallback) implementation (the same thing that we did for `fs.copyFile()`) and in that case we need to think about how we want to handle backwards compatibility when using `fs.cp()` and `fse.copy()`.

RyanZim:
`fs.cp` is still experimental, so holding off actually porting.

RyanZim:
As of Node v22.3.0, `fs.cp` is no longer experimental

RyanZim:
We haven't actually taken the time to do a `fs.cp` integration yet. However, later this year, Node v20 will reach EOL, so we may be able to just do a complete switch to `fs.cp`, without trying to match our implementation to Node's exactly (though they should be close).

I also need to ensure that a few fixes we've done since have also been applied to Node itself.

PR #1084, opened 2026-09-19
Respect overwrite options when copying symlinks

jakezwang:
Copying a symlink currently replaces an existing destination symlink even with `overwrite: false` or `clobber: false`. It also succeeds with `overwrite: false, errorOnExist: true` instead of reporting the existing destination.

Apply the existing overwrite checks in the symlink branches of `copy` and `copySync`, before unlinking the destination. Tests cover valid and broken destination links, both overwrite options, `errorOnExist`, default replacement and missing destinations.

The new tests reproduce 12 failures before the fix. `npm test` passes on Node 20.20.2, 22.23.2 and 24.4.1: lint, 755 unit tests and ESM tests. Eight existing tests remain pending on macOS; cross-device and other platform tests were not run locally.


RyanZim:
Of note, `fs.cp()` has the same behavior, and we would like to stay in sync with that: https://github.com/jprichardson/node-fs-extra/issues/912#issuecomment-891867716. So this should be potentially raised with Node itself.

jakezwang:
Understood. This was independently found edge-case behavior rather than a production incident. I have not opened an issue in Node, and I have not changed the proposed behavior further. Keeping `fs-extra` aligned with `fs.cp()` is a design constraint; this needs an upstream decision before treating the PR as ready to merge.

